PR #25, opened 2014-09-05
Bug 1063564 - l20n cause CSP violation for packaged App

gasolin:


stasm:
Fixed in bbb761d7681fef6c87f0068c58f16e11b769c467.  Thanks!

PR #29, opened 2015-01-04
Add the `default_locale` property in the manifest

Hywan:
Without any `default_locale` property, the `registerLocales` function is called with `(undefined, […])`. Because the first argument `defLocale` is undefined, the function returns without adding any available locales. So, the manifest is silently skipped.

### DIFF
--- a/docs/html.md
+++ b/docs/html.md
@@ -59,6 +59,7 @@ An example of the manifest file:
     
 ```json
 {
+  "default_locale": "en-US",
   "locales": [
     "en-US",
     "pl"


zbraniecki:
I don't think we use manifest anymore. @stasm - should we close this?

stasm:
v1.x used the manifest.  Later versions replaced it with `<meta>` tags.  The v1.x docs do mention `default_locale` (https://github.com/l20n/l20n.js/commit/a803a43e1e0b7db9a40b69600de7135daab7f237) so I think that we can close this.  Thanks for the pull request, @Hywan!

Hywan:
@stasm You're welcome :-].

PR #31, opened 2015-06-06
Move io to runtime

stasm:
I'd like to move io into runtime directories.  Here's how I started to think about this:

 - bindings/html is a set of modules required to localize HTML,
 - runtime/web/index is the exported api (window.L10n) in the web runtime,
 - io is specific to the runtime:  web has one, node has another, gaiabuild has yet another,
 - in the future, there should be runtime/gaiabuild/index which is the exported api for Gaia build system, with its io,
 - runtime/gaiabuild will make use of bindings/html to localize HTML,
 - and it will also make use of bindings/gaiabuild to perform the concatenation of the AST etc. (details TBD).

@zbraniecki, is this along the lines of how you've been thinking about the dir layout?


zbraniecki:
yeah, that sounds right.

PR #32, opened 2015-06-06
Introduce pretranslation

zbraniecki:
I'm pretty sure we can do this prettier, but it's the first try :)
@stasm - thoughts?

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -55,11 +55,15 @@ export function translateDocument(doc, langs) {
     dispatchEvent(doc, 'DOMLocalized', langs);
   };
 
-  doc.documentElement.lang = langs[0].code;
-  doc.documentElement.dir = langs[0].dir;
-  return translateFragment.call(
-    this, doc.documentElement).then(
-      setDOMLocalized, setDOMLocalized);
+  if (langs[0].code === doc.documentElement.getAttribute('lang')) {
+    return Promise.resolve(setDOMLocalized());
+  }
+
+  return translateFragment.call(this, doc.documentElement).then(() => {
+      doc.documentElement.lang = langs[0].code;
+      doc.documentElement.dir = langs[0].dir;
+      setDOMLocalized();
+  });
 }
 
 export function translateFragment(element) {
--- a/src/bindings/html/service.js
+++ b/src/bindings/html/service.js
@@ -17,6 +17,5 @@ export function initViews(langs) {
 function initView(view, langs) {
   dispatchEvent(view.doc, 'supportedlanguageschange', langs);
   return view.ctx.fetch(langs, 1).then(
-    translateDocument.bind(view, view.doc, langs)).then(
-      () => view.observe());
+    translateDocument.bind(view, view.doc, langs));
 }
--- a/src/bindings/html/view.js
+++ b/src/bindings/html/view.js
@@ -30,6 +30,8 @@ export class View {
     let observer = new MutationObserver(onMutations.bind(this));
     this.observe = () => observer.observe(this.doc, observerConfig);
     this.disconnect = () => observer.disconnect();
+
+    this.observe();
   }
 
   formatValue(id, args) {


src/bindings/html/langs.js
```diff
@@ -18,7 +18,7 @@ export function getAdditionalLanguages() {
 export function onlanguagechage(
   appVersion, defaultLang, availableLangs, requestedLangs) {
 
-  return this.languages = Promise.all([
+  return Promise.all([
```
stasm: There's a difference between assigning to `this.languages` here and in `changeLanguage`. Here `this.languages` is the new promise immediately after `onlanguagechange` is called.  In `changeLanguage` it's the old promise until `Promise.all()` is resolved asynchronously.  I think that would be a bug because `document.l10n.format*` would return translations in a non-deterministic language, depending on how long it takes for `Promise.all()` to resolve.

src/bindings/html/dom.js
```diff
@@ -49,17 +49,17 @@ function getTranslatables(element) {
...
 }
 
-export function translateDocument(doc, langs) {
-  let setDOMLocalized = function() {
-    doc.localized = true;
-    dispatchEvent(doc, 'DOMLocalized', langs);
-  };
+export function setDOMLocalized(langs) {
+  this.localized = true;
+  dispatchEvent(this, 'DOMLocalized', langs);
```
stasm: In other exported functions in this file, `this` is an instance of the `View`.  Would it be more consistent to keep it that way here and pass  `doc` as the first argument to `setDOMLocalized`?

src/bindings/html/langs.js
```diff
@@ -28,33 +28,50 @@ export function onlanguagechage(
...
-    initViews.call(this, langs);
+export function changeLanguage(
+  appVersion, defaultLang, availableLangs, additionalLangs, prevLangs,
+  requestedLangs) {
+
+  let newLangs = getLanguageChain(appVersion, defaultLang, availableLangs,
+    additionalLangs, prevLangs, requestedLangs);
+
+
+  if (prevLangs[0] === newLangs[0]) {
```
stasm: I think we should keep `arrEqual`.  In fallback scenarios, any change to the language chain regardless of the position in the chain should trigger the `supportedlanguageschange` event.

src/bindings/html/service.js
```diff
@@ -9,14 +9,18 @@ export const L10n = {
   change: null
 };
 
-export function initViews(langs) {
+export function initViews(pretranslated) {
```
stasm: Passing resolved `langs` to `initViews` helps avoid one additional `Promise.resolve`.  Why remove it?

src/bindings/html/dom.js
```diff
@@ -49,17 +49,17 @@ function getTranslatables(element) {
...
+}
 
+export function translateDocument(doc, langs) {
   doc.documentElement.lang = langs[0].code;
   doc.documentElement.dir = langs[0].dir;
   return translateFragment.call(
     this, doc.documentElement).then(
-      setDOMLocalized, setDOMLocalized);
+      setDOMLocalized.bind(doc, langs), setDOMLocalized.bind(doc, langs));
```
stasm: Maybe move this to `initView`?

src/runtime/web/index.js
```diff
@@ -40,9 +40,14 @@ function init() {
   this.views.push(
     document.l10n = new View(this, document));
 
+  var prevLangs = [];
+  if (document.documentElement.hasAttribute('lang')) {
```
stasm: Any logic directly related to the `document` should probably live in the view, not here.

src/bindings/html/langs.js
zbraniecki: Then, by default, if the document is pretranslated in 'fr', and the user has ['fr', 'en-US'], we will retranslate, which is imho a bug. Do you have a solution for that?

src/runtime/web/index.js
zbraniecki: not sure if I understand? why? lots of DOM related logic here.

src/bindings/html/langs.js
zbraniecki: ok, I fixed that (together with passing langs to initViews), but there's a bug in result, that I'd like to learn how to tackle in this code.

Basically, if we are in a pretranslated scenario, then we should not need `langs` in `changeLanguage`. We should be able to use prevLangs vs. newLangs to establish that we don't need changes, call `initViews` with pretranslated=true and do nothing else.

But because of the chain assignments in langs.on* and in `init`, we have to return langs, and we need to assign them to `this.languages`. How can we refactor that code to allow for the fastpath?

src/runtime/web/index.js
stasm: Each `document` should have its own `pretranslate` flag, local to the view that is managing this `document`.  I don't think it belongs here in runtime.

src/bindings/html/langs.js
stasm: I don't think we should be overloading `prevLangs` for the pretranslation optimization.

src/bindings/html/langs.js
stasm: So in general, I don't think we should be overloading `prevLangs` for the pretranslation optimization.  Could we instead simply check in `translateDocument` if there's already `lang` set on `documentElement` and if it's the same as `langs[0]` skip directly to `setDOMLocalized`?

src/bindings/html/langs.js
stasm: But if we end up doing that, there should also be a code path for changing user preferences that uses `arrEqual`.

src/runtime/web/index.js
stasm: Also, `prevLangs` is really the previously negotiated user's lang chain. I'm not sure it's the right abstraction for storing the language the `document` was pretranslated into.

src/bindings/html/service.js
```diff
@@ -16,7 +16,6 @@ export function initViews(langs) {
 
 function initView(view, langs) {
   dispatchEvent(view.doc, 'supportedlanguageschange', langs);
-  return view.ctx.fetch(langs, 1).then(
-    translateDocument.bind(view, view.doc, langs)).then(
-      () => view.observe());
+  view.ctx.fetch(langs, 1).then(
```
stasm: Please return the promise here.  Otherwise the `Promise.all` in `initViews` doesn't do anything useful.  (We're not taking advantage of it right now, but I think the code is cleaner with it.)

src/bindings/html/dom.js
```diff
@@ -55,11 +55,14 @@ export function translateDocument(doc, langs) {
...
 
-  doc.documentElement.lang = langs[0].code;
-  doc.documentElement.dir = langs[0].dir;
-  return translateFragment.call(
-    this, doc.documentElement).then(
-      setDOMLocalized, setDOMLocalized);
+  if (langs[0].code !== doc.documentElement.getAttribute('lang')) {
+    return translateFragment.call(this, doc.documentElement).then(() => {
+        doc.documentElement.lang = langs[0].code;
+        doc.documentElement.dir = langs[0].dir;
```
stasm: Maybe move these two assignments into `setDOMLocalized`?

src/bindings/html/dom.js
```diff
@@ -55,11 +55,14 @@ export function translateDocument(doc, langs) {
...
+        setDOMLocalized();
```
stasm: Should we also call `setDOMLocalized` if `translateFragment` errors out?  The old code did that, but I'm not sure if it's the right approach.  `Promise.all` in `translateFragment` could be rejected immediately when one of the child promises rejects, which sounds like not something we'd want.

src/bindings/html/dom.js
```diff
@@ -55,11 +55,14 @@ export function translateDocument(doc, langs) {
...
+    });
+  }
+  return Promise.resolve(setDOMLocalized());
```
stasm: It might be my stylistic preference, but I like putting the short-circuit code paths in `if`s.

stasm: …and the regular code path outside of the `if`.

src/bindings/html/dom.js
stasm: (Also, good call on moving them to *after* document translation is done!

zbraniecki: Can't because i need to call setDOMLocalized in pretranslated scenario

src/bindings/html/dom.js
zbraniecki: No, translateFragment error should not cause setDOMLocalized. It should never happen and not firing the event will make it easier to spot and debug

stasm: OK, this makes sense, thanks.

stasm:
I like how simple this turned out to be.  Thanks.

PR #33, opened 2015-06-06
V3 pending

zbraniecki:


src/bindings/html/dom.js
```diff
@@ -82,6 +88,11 @@ function camelCaseToDashed(string) {
 }
 
 export function translateElement(element) {
+  if (!this.service.languages) {
```
zbraniecki: agh, and the problem I hit here is that in non-pretranslated case, translateDocument is fired before service.languages is set :/

stasm:
Hmm, at the risk of sounding ignorant:  what is the purpose of `pendingElements` in v3 api which is always async?  I get that in v2 we needed to track sync `translateElement` calls made before the context was ready.  I think we might not need that at all in v3.  Perhaps one thing we could change is to call `view.observe()` on view initialization.  We disconnect the observer anyways when the translation is applied, so there's no risk of catching our own mutations.

zbraniecki:
that's a good point. The only question is, what happens when we fire formatValue (in translateElement) when service.languages is undefined? Can we somehow make it a promise that will resolve later?

stasm:
You shouldn't be able to do that because `this.languages` is created in https://github.com/l20n/l20n.js/blob/75768ae42a1bfcfd70c40394d8f231b8d0d870bb/src/runtime/web/index.js#L47, a mere two instructions after `document.l10n` is created.

stasm:
And both of those things are created in `init` which is run when `readyState` changes to `interactive`.  Any other code inserting elements into DOM should run after this, too, and after l20n.js.

stasm:
So, to recap:

    <script defer src="l20n.js"></script>
    <script defer src="inserThingstIntoDOM.js"></script>

We should document this as *the* way to plug l20n.js and other scripts into HTML.

zbraniecki:
you are right. We just need to start observing in View.init

stasm:
Great, that's less code to write! :)  Do you want to make this change in this PR?

zbraniecki:
yeah, I'll do this when we land the pretranslate

stasm:
Could we land this in parallel?  Up to you.

PR #34, opened 2015-06-08
De-this-ify

stasm:
I'd like to re-organize the code a bit to remove the uses of `this` outside of files which define classes to which `this` refers to.  Most importantly, `translate*` methods in dom.js take `view` as their first argument instead of implicitly being invoked with the view as `this`.  I also moved the service's `init` method into the service.js file.

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -50,7 +50,6 @@ function getTranslatables(element) {
 }
 
 export function translateDocument(ctx, obs, langs, doc) {
-  //XXX langs currently cannot be a promise 
   let setDOMLocalized = function() {
     doc.localized = true;
     dispatchEvent(doc, 'DOMLocalized', langs);
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -49,7 +49,8 @@ function getTranslatables(element) {
     nodes, element.querySelectorAll('*[data-l10n-id]'));
 }
 
-export function translateDocument(view, doc, langs) {
+export function translateDocument(ctx, obs, langs, doc) {
+  //XXX langs currently cannot be a promise 
   let setDOMLocalized = function() {
     doc.localized = true;
     dispatchEvent(doc, 'DOMLocalized', langs);
@@ -59,7 +60,7 @@ export function translateDocument(view, doc, langs) {
     return Promise.resolve(setDOMLocalized());
   }
 
-  return translateFragment(view, doc.documentElement).then(
+  return translateFragment(ctx, obs, langs, doc.documentElement).then(
     () => {
       doc.documentElement.lang = langs[0].code;
       doc.documentElement.dir = langs[0].dir;
@@ -67,10 +68,10 @@ export function translateDocument(view, doc, langs) {
     });
 }
 
-export function translateFragment(view, frag) {
+export function translateFragment(ctx, obs, langs, frag) {
   return Promise.all(
     getTranslatables(frag).map(
-      elem => translateElement(view, elem)));
+      elem => translateElement(ctx, obs, langs, elem)));
 }
 
 function camelCaseToDashed(string) {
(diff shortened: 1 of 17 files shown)


stasm:
@zbraniecki r?

src/bindings/html/dom.js
```diff
@@ -49,7 +49,7 @@ function getTranslatables(element) {
     nodes, element.querySelectorAll('*[data-l10n-id]'));
 }
 
-export function translateDocument(doc, langs) {
+export function translateDocument(view, doc, langs) {
```
zbraniecki: why do you need both, view and doc? You should be able to get doc out of view object, right?

src/bindings/html/service.js
```diff
@@ -1,21 +1,68 @@
...
+  appVersion, defaultLang, availableLangs, additionalLangs,
+  requestedLangs) {
+
+  return this.languages = this.languages.then(
+    prevLangs => changeLanguage(
+      initViews.bind(this), appVersion, defaultLang, availableLangs,
+      additionalLangs, prevLangs, requestedLangs || navigator.languages));
+}
+
+export function init(fetch, additionalLangsAtLaunch) {
```
zbraniecki: why this is not a method on L10n?

src/bindings/html/view.js
```diff
@@ -41,13 +43,21 @@ export class View {
...
 }
 
 View.prototype.setAttributes = setL10nAttributes;
 View.prototype.getAttributes = getL10nAttributes;
-View.prototype.translateFragment = translateFragment;
 
-function onMutations(mutations) {
+export function init(langs) {
+  dispatchEvent(this.doc, 'supportedlanguageschange', langs);
+  return translateDocument(this, this.doc, langs);
```
zbraniecki: I don't like this. View.init should not send supportedlanguageschange, it probably shouldn't even translateDocument, unless you have a flag to not do that. It feels like it doesn't belong here.

src/runtime/web/index.js
```diff
@@ -1,14 +1,8 @@
...
-import { L10n } from '../../bindings/html/service';
-import { View } from '../../bindings/html/view';
-import { getMeta } from '../../bindings/html/head';
-import {
-  changeLanguage, onlanguagechage, onadditionallanguageschange,
-  getAdditionalLanguages
-} from '../../bindings/html/langs';
+import { fetch } from './io';
+import { L10n, init } from '../../bindings/html/service';
+import { getAdditionalLanguages } from '../../bindings/html/langs';
```
zbraniecki: why are you loading getAdditionalLanguages here, just to pass it to service init?

zbraniecki:
ugh, I'm not sure what to think about those changes. Overall, I feel like there's some confusing overlap between bindings/html/service<-->runtime/web/index and bindings/html/view<--->bindings/html/dom, and I wanted to talk about considering flattening those, but the way you're moving the code around feels like it doesn't really benefit us and it makes each class abstraction less generic, or at least I don't understand the purpose.

src/bindings/html/service.js
stasm: I want this to be a privileged method: our code can call it, but the consumer of the API cannot, because it makes little sense to init the `L10n` singleton more than once.  This is why I define it here as an exported function and bind it to `L10n` in the runtime.

src/bindings/html/dom.js
stasm: Good point, I missed this.

src/bindings/html/view.js
stasm: Yeah, I'm not sure where `supportedlanguageschange` belongs.  If not here, than we don't need `View#init` at all, since all else it does is `translateDocument`.

src/runtime/web/index.js
stasm: I can call `getAdditionalLanguages` before `interactive` fires.  Service's `init` is run only after `interactive`.

stasm:
Thanks for the feedback! The back story here is that I started writing bindings/runtime for the Gaia's build system and noticed that in order to translate a document, I needed to create a service, an env and a view with its context.  I think for simpler scenarios it should be possible to just create a view and pass it the languages you want it to translate things into.

It all boils down the interface the `translate*` functions expect.  They can either expect a single `view`, or a set of the following arguments: context or the `formatEntity` method, languages and observer (or something that implements `observe` and `disconnect`).  The latter is more atomic and would make it easier to create other types of bindings.  The former requires the view to store the information about the languages, which may change.  I'd like to migrate towards the latter in this PR or in another one.

I'm also not sure if things like `View` and `Service` belong in bindings/html or perhaps in runtime/web.  Or perhaps in both, as a base class and a subclass, respectively.  I'd like to get your help with this;  I want to keep things simple and I'm afraid of over-engineering the solution.

stasm:
I continue to make changes which I find helpful in writing gaiabuild bindings.

src/bindings/html/view.js
```diff
@@ -35,44 +36,30 @@ export class View {
...
-        }
-      }
-    }
-
-    if (mutation.type === 'attributes') {
-      translateElement.call(this, mutation.target);
-    }
-  }
+export function translate(langs) {
+  dispatchEvent(this.doc, 'supportedlanguageschange', langs);
```
stasm: I'm considering removing `translate` entirely and instead calling `translateDocument` directly from the service, as well as not storing `doc` on `view` and instead keeping track of these relationships in a `WeakMap` in the service.

stasm:
OK, so I'm rather happy with where I ended up with this.  I cleaned up the code and moved certain functions to keep them logically closer to the object on which they are run.  Following is the summary of most notable changes:

 - I refactored the `translate*` methods to accept a context, an observer and an array of langs as arguments instead of a single view which would only obscure those three things.  In HTML bindings, the View happens to implement the observer interface (`observe` and `disconnect`) which is why I pass it verbatim, but it doesn't have to be that way.

 - I tried to use more fat arrow functions _in lieu_ of `Function.prototype.bind` where possible.  Fat arrow functions are faster and consume less memory.  I might have missed a few occasions.

 - I greatly simplified the `Service` class (which is now a class, not a singleton) and separated the initialization from the attaching of event handlers on `window` and `document`.

 - I made the `langs` argument passed to `ctx` and `translate*` methods consistent.  Previously there was a bit of dichotomy here:  you could pass an array or a promise resolving to an array.  I now require a regular array so the promise needs to be resolved earlier by the calling code.  Which works really well with the `View` abstraction now!

 - I removed most of the default exports because I caught myself constantly checking if a given module has default or named exports.  Named are explicit and easier to extend in the future.  I only left default exports in lib/format since that's your department, @zbraniecki, but I'll be happy to do the same there, or leave that for another PR.

There are no business logic changes in this PR, it's all cleaning up and moving things around.

Let me know what you think, I'd love to land this soon and move on the Gaia buildtime bindings :)

src/bindings/html/view.js
stasm: This is the only place in which a `view` uses its `doc`.

zbraniecki:
1) Why is Service a class now?
2) I thought that View is a public interface to Context, so gaia build will have it's own public interface instead of exposing translate*

I'll review it today

stasm:
1. It's consistent with other abstraction that way and can be instantiated in a more flexible manner. Also, I can imagine multi-language apps which might actually require more than one `Service`.

2. That's right, gaiabuild will have its own variation of a View-like class:  without observer and without service.  You'll be able to pass language directly to it instead of storing it as a promise in view.service.languages.

src/bindings/html/view.js
zbraniecki: fine with me

stasm: I'll create a new PR for this.

PR #37, opened 2015-06-15
Turn off MutationObserver while translating DOM

zbraniecki:
Didn't measure perf/mem on this yet

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -73,7 +73,7 @@ export function translateMutations(view, langs, mutations) {
   for (let mutation of mutations) {
     switch (mutation.type) {
       case 'attributes':
-        translateElement(view, langs, mutation.target);
+        targets.add(mutation.target);
         break;
       case 'childList':
         for (let addedNode of mutation.addedNodes) {
@@ -81,19 +81,29 @@ export function translateMutations(view, langs, mutations) {
             targets.add(addedNode);
           }
         }
+        break;
     }
   }
 
-  targets.forEach(
-    target => target.childElementCount ?
-      translateFragment(view, langs, target) :
-      translateElement(view, langs, target));
+  if (targets.size === 0) {
+    return;
+  }
+
+  let elements = [];
+
+  targets.forEach(target => target.childElementCount ?
+      elements.concat(getTranslatables(target)) : elements.push(target));
+
+  Promise.all(
+    elements.map(elem => getElementTranslation(view, langs, elem))).then(
+      translations => applyTranslations(view, elements, translations));
 }
 
 export function translateFragment(view, langs, frag) {
+  let elements = getTranslatables(frag);
   return Promise.all(
-    getTranslatables(frag).map(
-      elem => translateElement(view, langs, elem)));
+    elements.map(elem => getElementTranslation(view, langs, elem))).then(
+      translations => applyTranslations(view, elements, translations));
 }
 
 function camelCaseToDashed(string) {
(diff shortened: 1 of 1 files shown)


stasm:
It would be interesting to see if this improves perf.  If it does, we'd know that we're paying for the many disconnects of the mutation observer.  And if we do, I'm afraid that we'll need to look for saving somewhere else.

I tried a similar approach as well.  I tried disconnecting in `onMutations` and then waiting for all promises returned by `translate*` methods there and reconnecting.  I realized however that this wouldn't be a valid approach.  Please correct me if I misunderstand:

Event loop pass 1:

 - a mutation happens and `onMutations` is called,
 - in it, or in `translateElement`, the observer is disconnected,
 - `translateElement` asynchronously asks the `ctx` for translation.
 
Event loop pass 2:

 - developer inserts a new element into the DOM,
 - the observer is disconnected; the new element is not translated.

Event loop pass 3:

 - the `ctx` returns the translations and resolves the `translateElement` promise,
 - the translation is applied,
 - the observer is re-connected.

The thing is, we can't globally disconnect and re-connect the observer if there's any async operation happening in between.  We need to pass the `view` down to the last async function call and disconnect and re-connect it synchronously.

zbraniecki:
could we reuse pendingElements when observer is disconnected?

stasm:
How would we know which element to add to `pending` if the observer is disconnected?

zbraniecki:
yeah, that didn't make much sense. Another attempt. Thoughts?

src/bindings/html/dom.js
```diff
@@ -116,13 +139,25 @@ export function translateElement(view, langs, elem) {
     return false;
   }
 
-  return view.ctx.formatEntity(langs, l10n.id, l10n.args).then(
-    translation => applyTranslation(view, elem, translation));
+  return view.ctx.formatEntity(langs, l10n.id, l10n.args).then(translation => {
+    view.disconnect();
```
stasm: Why do you disconnect here *and* inside of `applyTranslation`?

stasm: Ah, ok, I see.  The diff colors misled me.

src/bindings/html/dom.js
```diff
@@ -81,19 +81,32 @@ export function translateMutations(view, langs, mutations) {
...
         }
+        break;
     }
   }
 
-  targets.forEach(
-    target => target.childElementCount ?
-      translateFragment(view, langs, target) :
-      translateElement(view, langs, target));
+  if (targets.size) {
```
stasm: Can you return early here if `targets.size` is `0`?

src/bindings/html/dom.js
```diff
@@ -81,19 +81,32 @@ export function translateMutations(view, langs, mutations) {
...
+      if (target.childElementCount) {
+        elements.concat(getTranslatables(target));
+      } else {
+        elements.push(target);
+      }
+    });
+
+    Promise.all(
+      elements.map(elem => getElementTranslation(view, langs, elem))).then(
+        (translations) => applyTranslations(view, elements, translations));
```
stasm: No need for `(` and `)` around `translations` here, I think.

src/bindings/html/dom.js
```diff
@@ -109,6 +122,16 @@ function camelCaseToDashed(string) {
...
 }
 
+function getElementTranslation(view, langs, elem) {
+  var l10n = getAttributes(elem);
+
+  if (!l10n.id) {
+    return false;
+  }
+
+  return view.ctx.formatEntity(langs, l10n.id, l10n.args);
```
stasm: Maybe make this shorter with a ternary if?  `return l10n.id ? ...`

src/bindings/html/dom.js
```diff
@@ -116,13 +139,25 @@ export function translateElement(view, langs, elem) {
     return false;
   }
 
-  return view.ctx.formatEntity(langs, l10n.id, l10n.args).then(
-    translation => applyTranslation(view, elem, translation));
+  return view.ctx.formatEntity(langs, l10n.id, l10n.args).then(translation => {
```
stasm: Should this use `getElementTranslation`?

zbraniecki:
r?

src/bindings/html/dom.js
```diff
@@ -81,19 +81,34 @@ export function translateMutations(view, langs, mutations) {
...
-    target => target.childElementCount ?
-      translateFragment(view, langs, target) :
-      translateElement(view, langs, target));
+  if (targets.size === 0) {
+    return;
+  }
+
+  let elements = [];
+
+  targets.forEach(target => {
```
stasm: One last stylistic change, please indulge me :) I find the x => { } syntax difficult to read.  Can we make this snippet more lambda-like:

    targets.forEach(target => target.childElementCount ?
      elements.concat(getTranslatables(target) : elements.push(target));

This also has the benefit of being shorter :)

zbraniecki: ugh, sure. But I hope you'll get familiar with multi-statement fat arrow functions one day! We can't turn them all into lambdas.

stasm: Yeah, I know. Still, there's something about syntax like https://github.com/l20n/l20n.js/blob/cbd9c5162833fe5d84e4224ea28b4f016d231760/src/bindings/html/dom.js#L63-L67 which I've yet to get used to.

PR #39, opened 2015-06-16
Delay mozApps.getAdditionalLanguages

stasm:
The numbers at https://l20n.etherpad.mozilla.org/v3-perf suggest that it would be beneficial to delay the `mozApps.getAdditionalLanguages` call after all.  Also, this is what master's l10n.js currently does as well, so we might want to keep the parity, at least for now.


stasm:
@zbraniecki What's your take?

zbraniecki:
let's land it for the sake of parity. I want to minimize the differences that affect performance until we get on par.

PR #42, opened 2015-06-29
V3 runtime parser

zbraniecki:
@stasm - f?

PR #43, opened 2015-06-30
Remove the meta WeakMap

stasm:
I was reviewing #42 and I realized that we don't actually need to store the `id` and `lang` for each entity.  The `id` is needed for the error recovery which can be (and even should be) done in the context, and `lang` is needed for getting the right values for placeables and macros, but it should be constant during a single formatting operation.

I decided to go ahead and suggest this change.  @zbraniecki, I can help you rebase #42 on top of this, I'm expecting some (but not many) conflicts.

### DIFF
--- a/src/lib/resolver.js
+++ b/src/lib/resolver.js
@@ -12,10 +12,9 @@ var nonLatin1 = /[^\x01-\xFF]/;
 var FSI = '\u2068';
 var PDI = '\u2069';
 
-const meta = new WeakMap();
 const resolutionChain = new WeakSet();
 
-export function createEntry(node, lang) {
+export function createEntry(node) {
   var keys = Object.keys(node);
 
   // the most common scenario: a simple string with no arguments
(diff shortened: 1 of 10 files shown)


stasm:
I don't see any change in raptor tests on FM using French as the non-default language.  The code gets simpler, however, so: @zbraniecki r?

src/lib/resolver.js
```diff
@@ -34,42 +33,35 @@ export function createEntry(node, lang) {
...
-export function format(ctx, args, entity) {
+export function format(ctx, lang, args, entity) {
   if (typeof entity === 'string') {
     return [{}, entity];
   }
 
   if (resolutionChain.has(entity)) {
-    const m = meta.get(entity);
-    throw new L10nError('Cyclic reference detected: ' + m.id, m.id, m.lang);
+    throw new L10nError('Cyclic reference detected');
```
zbraniecki: is this the only thing we are loosing? Can we at least get add a lang then?

zbraniecki:
r+

src/lib/resolver.js
stasm: We'll catch the error in the context and add the id and the lang.

stasm: The only thing we're losing is that this error might happen in a different entity than the currently formatted one, and `err.id` will be the name of the latter.

PR #44, opened 2015-07-01
Move resolutionChain into locals

stasm:
By moving the resolution chain into locals we're reducing the complexity in `format`: we can get rid of one `try...finally`.  Since we're creating a new WeakSet with every `format` call, this is a tiny tiny bit slower:  `grunt perf` reports 300 microseconds slower, or 12%.

### DIFF
--- a/src/bindings/gaiabuild/view.js
+++ b/src/bindings/gaiabuild/view.js
@@ -50,8 +50,8 @@ export class View {
     }
   }
 
-  observe() {}
-  disconnect() {}
+  _observe() {}
+  _disconnect() {}
 
   _resolveEntities(langs, keys) {
     return this.ctx.resolveEntities(langs, keys);
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -88,9 +88,9 @@ function translateElements(view, langs, elements) {
 }
 
 function applyTranslations(view, elems, translations) {
-  view.disconnect();
+  view._disconnect();
   for (let i = 0; i < elems.length; i++) {
     overlayElement(elems[i], translations[i]);
   }
-  view.observe();
+  view._observe();
 }
--- a/src/bindings/html/view.js
+++ b/src/bindings/html/view.js
@@ -16,10 +16,10 @@ const observerConfig = {
 
 export class View {
   constructor(client, doc) {
-    this.doc = doc;
+    this._doc = doc;
     this.qps = qps;
 
-    this.interactive = documentReady().then(
+    this._interactive = documentReady().then(
       () => init(this, client));
 
     this.ready = new Promise(function(resolve) {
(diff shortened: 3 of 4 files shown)


stasm:
@zbraniecki r?

I know you'd rather remove `locals` completely, but I think that we'll need them soon for the `ThisExpression`. I don't see how to implement this with a global `Weak*`:  we need a stack from which we can take the tail value for know which entity `ThisExpression` points to.

I'll keep thinking about this.  Technically, we could remove locals right now since we don't support `ThisExpression` yet, but it's just a lot of work to go back and forth on this, especially in tests.

zbraniecki:
Do you need this patch anytime soon to be landed?
can we wait a little bit with this? I'd like to think and chat with you more about the role of locals and the resolutionChain.

stasm:
No, not needed right now, it's just the next step after the meta week map pr.

askmatey:
By introducing **ES6** `WeakSet`, the latest `3.x` l20n builds are not compatible with Firefox OS `<=2.0` anymore. 

See https://wiki.mozilla.org/Releases#Based_on_Gecko_38 :

> ...
> Firefox OS 2.0 - Based on Gecko `32`
> Firefox OS 2.1 - Based on Gecko `34`

browser-comp-weakset

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/WeakSet#Browser_compatibility

zbraniecki:
Correct. That was intentional. Is there a question in your statement?

stasm:
I'd like to revisit this pull request.  `WeakSet` support is ATM worse than `WeakMap` for instance, so I've started considering replacing it with an array.  Having the array in `locals` would be perfectly fine.  It would help reduce the amount of polyfills l20n.js needs and make the code in `format` simpler.  @zbraniecki, what's your take on this?

zbraniecki:
So, overall I'm ok with this change, but I'd prefer not to have to do this.

One question is - can we fix it with a polyfill - https://github.com/dfcreative/weakset ?

stasm:
We could, but my point about polyfills still stands:  I suggest we resort to polyfills for features that are hard or nearly impossible to work around: Promise and the <template> element are good examples.  Here, there's an easy way around this and I'd be afraid of requiring people to use yet another polyfill to get our code working. I think two is already a lot :(

askmatey:
Native ES6 is fast, do not get me wrong. The usage of `WeakSet` in this case is justifiable and understandable. But any additional shim's affect memory usage, especially on low-memory devices like Firefox OS.

Just to give you an example:
By adding the `WeakSet` pollyfill from [zloirock/core-js](https://github.com/zloirock/core-js) project, the general memory usage of my Firefox OS app grew around 2MB. 

However, that's all right with me. I do not want to prevent browser engines from using their own optimized ES6 implementations and I prefer to use a mix between ES5 and ES6 functions by targeting specific environments as you do in this project. 

The [ECMAScript compatibility table](https://kangax.github.io/compat-table/es6/) gives you an excellent overview.

```js
if (typeof window.WeakSet === 'undefined') {
  require('core-js/es6/weak-set');
}
```

stasm:
Closing this as it applies only to the v3.x resolver and isn't big enough an improvement to really consider landing it there.

PR #45, opened 2015-07-01
Gaiabuild

stasm:
My work in https://bugzil.la/1169361 is not done yet because of Gij failures, but I'd like to merge this in already here to avoid bitrot. The biggest conceptual change is that `translateDocument` is now a method of the `View` class which allows me to apply different logic on runtime and buildtime.

@zbraniecki r?


stasm:
I also added `view.emit` to keep other parts of bindings environment-agnostic. E.g. we used to `console.warn` in dom.js's `applyTranslation`, but `console` is not defined in buildtime.  I changed that to emit a `deprecatewarning` via `view.emit`.

stasm:
I now have a [green build](https://treeherder.mozilla.org/#/jobs?repo=gaia&revision=0fba923e219f70d89ac62393b2d85e52542e50aa) using this branch and changes to webapp-optimize from https://github.com/mozilla-b2g/gaia/pull/30368. It all works \o/

src/bindings/gaiabuild/view.js
```diff
@@ -0,0 +1,120 @@
+'use strict';
+
+import { Env } from '../../lib/env';
+import { qps } from '../../lib/pseudo';
+import { getResourceLinks } from '../../bindings/html/head';
+import { translateFragment } from '../../bindings/html/dom';
+import { getDirection } from '../../bindings/html/langs';
+import { serializeContext } from '../../bindings/gaiabuild/serialize';
```
zbraniecki: why do you use '../../' instead of './serialize' etc?

src/bindings/gaiabuild/view.js
```diff
@@ -0,0 +1,120 @@
...
+    this.env.addEventListener('*', amendError.bind(this));
+
+    this.stopBuildError = null;
+    const log = logError.bind(this);
+    const stop = stopBuild.bind(this);
+
+    // stop the build if these errors happen for en-US
+    // XXX tv_apps break the build https://bugzil.la/1179833
+    // this.env.addEventListener('fetcherror', stop);
+    // XXX parse errors don't have the lang property needed by stopBuild
```
zbraniecki: Then I'd argue that we should intercept the error, add a lang and push it forward instead of transforming parser API to be aware of the lang it's parsing. Parser should be dumb imho.

src/bindings/html/head.js
```diff
@@ -3,7 +3,7 @@
 export function getResourceLinks(head) {
   return Array.prototype.map.call(
     head.querySelectorAll('link[rel="localization"]'),
-    (el) => el.getAttribute('href'));
+    el => decodeURI(el.getAttribute('href')));
```
zbraniecki: why do you need decodeURI here?

src/lib/events.js
```diff
@@ -3,13 +3,13 @@
...
+    listeners['*'].slice().forEach(
       listener => listener.apply(this, args));
   }
 
-  if (listeners['*']) {
-    listeners['*'].slice().forEach(
+  if (listeners[type]) {
+    listeners[type].slice().forEach(
       listener => listener.apply(this, args));
   }
```
zbraniecki: why did you reverse the order here?

zbraniecki:
r+

src/bindings/html/head.js
stasm: Some apps have own build scripts which use `nsIDOMSerializer` to serialize the HTML, which escapes `{}` in `href` attributes.  We ran into this in the Keyboard app in https://bugzil.la/1098188.

src/bindings/gaiabuild/view.js
stasm: Yeah, I agree with you although I'm not sure how exactly the intercepting should look like.  I'm fine leaving this to be figured out at a later date;  we don't do this in l10n.js right now, so we don't need to introduce this new error report right now.

src/bindings/gaiabuild/view.js
stasm: Probably a copy-paste gone wrong.  Thanks for catching it!

src/lib/events.js
stasm: I'm using the `*` event listener to add the name of the current app in `amendError`. I need it to run before any other event handlers, so that if any of them logs the error directly to stdout, it already has the amended message.

I'm sure there's a better way of achieving this, but this seemed easy and harmless enough :)  Any ideas?

PR #46, opened 2015-07-02
Use const and let

stasm:
Let's do this.

(This is very easy to rebase, so I'm fine waiting for #42 first.)


stasm:
@zbraniecki r?

PR #47, opened 2015-07-02
Pass an emitter function to parsers

stasm:
We only need to pass the `env` into `parse` because of error emitting.  Let's just pass the emitter function instead, which gives us greater flexibility in terms of what that function actually is.  E.g. we can add meta data to emitted errors which is not available inside of the parser (like the current `lang`).


stasm:
I checked tests and tools/* scripts and it looks like we're mostly passing `null` in those.

stasm:
@zbraniecki r?

PR #48, opened 2015-07-03
Introduce legacy mode for gaiabuild bindings 

stasm:
The buildtime view can now optionally use the v2 parser, createEntry and serializer for apps which use shared/js/l10n.js.


stasm:
@zbraniecki r?

PR #49, opened 2015-07-07
V3 runtime parser

zbraniecki:
I still have some tweaks to do here and there, but it passes tests and lint. @stasm - r?

### DIFF
--- a/src/bindings/gaiabuild/legacy/env.js
+++ b/src/bindings/gaiabuild/legacy/env.js
@@ -4,7 +4,7 @@ import { L10nError } from '../../../lib/errors';
 import { Env, amendError } from '../../../lib/env';
 import { createEntry } from './resolver';
 import PropertiesParser from './parser';
-import { walkContent, qps } from '../../../lib/pseudo';
+import { walkEntry, qps } from '../../../lib/pseudo';
 
 export class LegacyEnv extends Env {
   _parse(syntax, lang, data) {
@@ -31,6 +31,5 @@ export class LegacyEnv extends Env {
 }
 
 function createPseudoEntry(node, lang) {
-  return createEntry(
-    walkContent(node, qps[lang.code].translate), lang);
+  return createEntry(walkEntry(node, qps[lang.code].translate));
 }
--- a/src/lib/env.js
+++ b/src/lib/env.js
@@ -1,11 +1,9 @@
 'use strict';
 
-import { L10nError } from './errors';
 import { Context } from './context';
-import { createEntry } from './resolver';
 import PropertiesParser from './format/properties/parser';
 import L20nParser from './format/l20n/parser';
-import { walkContent, qps } from './pseudo';
+import { walkEntry, qps } from './pseudo';
 import { emit, addEventListener, removeEventListener } from './events';
 
 const parsers = {
@@ -40,23 +38,6 @@ export class Env {
     return parser.parse.call(parser, emit, data);
   }
 
-  _createEntries(lang, ast) {
-    const entries = Object.create(null);
-    const create = lang.src === 'qps' ?
-      createPseudoEntry : createEntry;
-
-    for (let i = 0, node; node = ast[i]; i++) {
-      const id = node.$i;
-      if (id in entries) {
-        this.emit('duplicateerror', new L10nError(
-         'Duplicate string "' + id + '" found in ' + lang.code, id, lang));
-      }
-      entries[id] = create(node, lang);
-    }
-
-    return entries;
-  }
-
   _getResource(lang, res) {
     const cache = this._resCache;
     const id = res + lang.code + lang.src;
@@ -68,8 +49,15 @@ export class Env {
     const syntax = res.substr(res.lastIndexOf('.') + 1);
 
     const saveEntries = data => {
-      const ast = this._parse(syntax, lang, data);
-      cache[id] = this._createEntries(lang, ast);
+      const entries = this._parse(syntax, lang, data);
+      if (lang.src !== 'qps') {
+        cache[id] = entries;
+      } else {
+        cache[id] = Object.create(null);
+        for (let key in entries) {
+          cache[id][key] = walkEntry(entries[key], qps[lang.code].translate);
+        }
+      }
     };
 
     const recover = err => {
@@ -85,11 +73,6 @@ export class Env {
     return cache[id] = this.fetch(res, langToFetch).then(
       saveEntries, recover);
   }
-}
-
-function createPseudoEntry(node, lang) {
-  return createEntry(
-    walkContent(node, qps[lang.code].translate), lang);
 }
 
 export function amendError(lang, err) {
--- a/src/lib/format/l20n/parser.js
+++ b/src/lib/format/l20n/parser.js
@@ -5,183 +5,122 @@ import { L10nError } from '../../errors';
 const MAX_PLACEABLES = 100;
 
 export default {
-  _patterns: null,
-
-  init: function() {
-    this._patterns = {
-      index: /@cldr\.plural\(\$?(\w+)\)/g,
-      placeables: /\{\{\s*\$?([^\s]*?)\s*\}\}/,
-      unesc: /\\({{|u[0-9a-fA-F]{4}|.)/g,
-    };
-  },
-
-  parse: function (emit, string, simple) {
-    if (!this._patterns) {
-      this.init();
-    }
+  parse: function(emit, string) {
     this._source = string;
     this._index = 0;
-    this._length = this._source.length;
-    this.simpleMode = simple;
+    this._length = string.length;
+    this.entries = Object.create(null);
     this.emit = emit;
 
-    return this.getL20n();
+    return this.getResource();
   },
 
-  getAttributes: function() {
-    let attrs = Object.create(null);
-
-    while (true) {
-      let attr = this.getKVPWithIndex();
-      attrs[attr[0]] = attr[1];
-      const ws1 = this.getRequiredWS();
-      const ch = this._source.charAt(this._index);
-      if (ch === '>') {
-        break;
-      } else if (!ws1) {
-        throw this.error('Expected ">"');
-      }
-    }
-    return attrs;
-  },
-
-  getKVP: function() {
-    const key = this.getIdentifier();
+  getResource: function() {
     this.getWS();
-    if (this._source.charAt(this._index) !== ':') {
-      throw this.error('Expected ":"');
-    }
-    ++this._index;
-    this.getWS();
-    return [key, this.getValue()];
-  },
-
-  getKVPWithIndex: function() {
-    const key = this.getIdentifier();
-    let index = null;
-
-    if (this._source.charAt(this._index) === '[') {
-      ++this._index;
-      this.getWS();
-      index = this.getIndex();
-    }
-    this.getWS();
-    if (this._source.charAt(this._index) !== ':') {
-      throw this.error('Expected ":"');
-    }
-    ++this._index;
-    this.getWS();
-    return [
-      key,
-      this.getValue(false, undefined, index)
-    ];
-  },
-
-  getHash: function() {
-    ++this._index;
-    this.getWS();
-    let hash = {};
-    while (true) {
-      const hi = this.getKVP();
-      hash[hi[0]] = hi[1];
-      this.getWS();
-
-      const comma = this._source.charAt(this._index) === ',';
-      if (comma) {
-        ++this._index;
-        this.getWS();
-      }
-      if (this._source.charAt(this._index) === '}') {
-        ++this._index;
-        break;
-      }
-      if (!comma) {
-        throw this.error('Expected "}"');
-      }
-    }
-    return hash;
-  },
-
-  unescapeString: function(str, opchar) {
-    function replace(match, p1) {
-      switch (p1) {
-        case '\\':
-          return '\\';
-        case '{{':
-          return '{{';
-        case opchar:
-          return opchar;
-        default:
-          if (p1.length === 5 && p1.charAt(0) === 'u') {
-            return String.fromCharCode(parseInt(p1.substr(1), 16));
+    while (this._index < this._length) {
+      try {
+        this.getEntry();
+      } catch (e) {
+        if (e instanceof L10nError) {
+          // we want to recover, but we don't need it in entries
+          this.getJunkEntry();
+          if (!this.emit) {
+            throw e;
           }
-          throw this.error('Illegal unescape sequence');
-      }
-    }
-    return str.replace(this._patterns.unesc, replace.bind(this));
-  },
-
-  getString: function(opchar) {
-    let opcharPos = this._source.indexOf(opchar, this._index + 1);
-
-    outer:
-    while (opcharPos !== -1) {
-      let backtrack = opcharPos - 1;
-      // 92 === '\'
-      while (this._source.charCodeAt(backtrack) === 92) {
-        if (this._source.charCodeAt(backtrack - 1) === 92) {
-          backtrack -= 2;
         } else {
-          opcharPos = this._source.indexOf(opchar, opcharPos + 1);
-          continue outer;
+          throw e;
         }
       }
-      break;
+
+      if (this._index < this._length) {
+        this.getWS();
+      }
     }
 
-    if (opcharPos === -1) {
-      throw this.error('Unclosed string literal');
-    }
-
-    let buf = this._source.slice(this._index + 1, opcharPos);
-
-    this._index = opcharPos + 1;
-
-    if (!this.simpleMode && buf.indexOf('\\') !== -1) {
-      buf = this.unescapeString(buf, opchar);
-    }
-
-    if (!this.simpleMode && buf.indexOf('{{') !== -1) {
-      return this.parseString(buf);
-    }
-
-    return buf;
+    return this.entries;
   },
 
-  getValue: function(optional, ch, index) {
-    let val;
-
-    if (ch === undefined) {
-      ch = this._source.charAt(this._index);
-    }
-    if (ch === '\'' || ch === '"') {
-      val = this.getString(ch);
-    } else if (ch === '{') {
-      val = this.getHash();
-    }
-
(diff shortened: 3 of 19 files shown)


src/lib/env.js
```diff
@@ -40,21 +38,12 @@ export class Env {
...
-         'Duplicate string "' + id + '" found in ' + lang.code, id, lang));
-      }
-      entries[id] = create(node, lang);
+  _createEntries(lang, entries) {
+    if (lang.src !== 'qps') {
+      return entries;
     }
 
-    return entries;
+    return entries.map(createPseudoEntry.bind(this, lang));
```
stasm: We should be able to simplify this even more by calling `walkContent(entries, ...)`.

src/lib/env.js
```diff
@@ -40,21 +38,12 @@ export class Env {
...
-  _createEntries(lang, ast) {
-    const entries = Object.create(null);
-    const create = lang.src === 'qps' ?
-      createPseudoEntry : createEntry;
-
-    for (let i = 0, node; node = ast[i]; i++) {
-      const id = node.$i;
-      if (id in entries) {
-        this.emit('duplicateerror', new L10nError(
-         'Duplicate string "' + id + '" found in ' + lang.code, id, lang));
```
stasm: Please move duplicates detection into the parser.

src/lib/format/l20n/parser.js
```diff
@@ -5,201 +5,126 @@ import { L10nError } from '../../errors';
...
-  parse: function (emit, string, simple) {
-    if (!this._patterns) {
-      this.init();
-    }
+  parse: function(env, string) {
     this._source = string;
     this._index = 0;
-    this._length = this._source.length;
-    this.simpleMode = simple;
-    this.emit = emit;
```
stasm: Why are you removing `emit`?

src/lib/format/l20n/parser.js
```diff
@@ -5,201 +5,126 @@ import { L10nError } from '../../errors';
...
-  getAttributes: function() {
-    let attrs = Object.create(null);
+  getResource: function() {
+    this.getWS();
+    while (this._index < this._length) {
+      try {
+        this.getEntry();
+      } catch (e) {
+        if (e instanceof L10nError) {
+          throw e;
```
stasm: I think we'll want to `emit` here, but if not, maybe get rid of the `try...catch`?

src/lib/format/l20n/parser.js
```diff
@@ -5,201 +5,126 @@ import { L10nError } from '../../errors';
...
     this._source = string;
     this._index = 0;
-    this._length = this._source.length;
-    this.simpleMode = simple;
-    this.emit = emit;
+    this._length = string.length;
+    this.entries = Object.create(null);
 
-    return this.getL20n();
+    return this.getResource();
```
stasm: I wonder if this could be a good opportunity to refactor the parser a bit by not using `this` at all. Right now, invoking the `parse` method requires you to bind it to the `*Parser` object.  What if we stored `source`, `index`, `length` and `entries` in a new object , `state`, created when `parse` is called, and passed it into every function: `getResource(state)`?

src/lib/format/l20n/parser.js
```diff
@@ -224,153 +149,332 @@ export default {
...
-        if (this.emit) {
-          this.emit('parseerror', e);
-        } else {
-          throw e;
+
+    let defKey;
+
+    while (true) {
+
+
```
stasm: Whitespace?

src/lib/format/l20n/parser.js
```diff
@@ -224,153 +149,332 @@ export default {
...
+        index
+      };
     }
-    throw this.error('Invalid entry');
   },
 
-  getL20n: function() {
-    const ast = [];
+  getHash: function() {
+    let items = Object.create(null);
```
stasm: Use `const`here.

src/lib/format/l20n/parser.js
```diff
@@ -224,153 +149,332 @@ export default {
...
-          this.emit('parseerror', e);
-        } else {
-          throw e;
+
+    let defKey;
+
+    while (true) {
+
+
+      let [key, value, def] = this.getHashItem();
```
stasm: `const`

src/lib/format/l20n/parser.js
```diff
@@ -224,153 +149,332 @@ export default {
...
+      items[key] = value;
+
+      if (def) {
+        if (defKey) {
+          throw this.error('Default item redefinition forbidden');
         }
+        defKey = key;
```
stasm: Maybe use `var defKey = key` here instead of declaring `let defKey` above?

src/lib/format/l20n/parser.js
```diff
@@ -224,153 +149,332 @@ export default {
...
+  },
+
+  getAttribute: function(attrs) {
+    const key = this.getIdentifier();
+    let index;
+
+    if (this._source[this._index]=== '[') {
+      ++this._index;
+      this.getWS();
+      index = this.getItemList(this.getExpression, ']');
```
stasm: `var index = ...`?

src/lib/format/l20n/parser.js
```diff
@@ -224,153 +149,332 @@ export default {
...
+      cmpt: computed
+    };
+  },
+
+  getCallExpression: function(callee) {
+    this.getWS();
+
+    return {
+      type: 'call',
+      call: callee,
```
stasm: I know I was the one who suggested it, but I think `call` is confusing because of `Function.prototype.call`.  Maybe `expr`, similar to `PropertyExpression`?

src/lib/format/l20n/parser.js
```diff
@@ -224,153 +149,332 @@ export default {
...
+
+  error: function(message) {
+    const pos = this._index;
+
     let start = this._source.lastIndexOf('<', pos - 1);
-    const lastClose = this._source.lastIndexOf('>', pos - 1);
+    let lastClose = this._source.lastIndexOf('>', pos - 1);
     start = lastClose > start ? lastClose + 1 : start;
-    const context = this._source.slice(start, pos + 10);
+    let context = '\x1b[90m' + this._source.slice(start, pos) + '\x1b[0m';
```
stasm: What's this? Colors?

stasm:
Great work, r=me with the following comments:

 - re-add duplicates detection in the parser,
 - use `const` where possible, then `let`, then `var`,
 - fix tools/{parse,format}.

src/lib/env.js
zbraniecki: umm, walkContent seems to take a single entry. Do you want me to turn it around?

src/lib/env.js
zbraniecki: I did. But that doesn't solve the scenario in which two entities overlap between resources (in non-.json mode) in one context. Should we skip it for now?

src/lib/format/l20n/parser.js
zbraniecki: I'm willing to give your approach a try, although I suspect it will cost more.
But I'd prefer not to mix this refactor into this patch. I'll take it as a separate one after this, ok?

src/lib/format/l20n/parser.js
zbraniecki: forgot to bring it back. thanks for catching this!

src/lib/format/l20n/parser.js
zbraniecki: I believe that would make the code less readable (guessing scope hoisting)

zbraniecki:
Cool! I addressed your feedback. Can you look at the last commit in this PR and tell me if it looks good? I'll squash and merge.

src/lib/format/l20n/parser.js
stasm: OK.

src/lib/format/l20n/parser.js
stasm: Absolutely, makes sense.

src/lib/env.js
stasm: I think it takes any object and recursively walks over its properties.

src/lib/env.js
```diff
@@ -68,8 +49,13 @@ export class Env {
...
 
     const saveEntries = data => {
-      const ast = this._parse(syntax, lang, data);
-      cache[id] = this._createEntries(lang, ast);
+      const entries = this._parse(syntax, lang, data);
+      if (lang.src !== 'qps') {
+        cache[id] = entries;
+      } else {
+        cache[id] = entries.map(entry =>
+          walkEntry(entry, qps[lang.code].translate));
```
stasm: `entries` being an object, it doesn't have a `map` method.

src/lib/pseudo.js
```diff
@@ -1,27 +1,45 @@
 'use strict';
 
-// Recursively walk an AST node searching for content leaves
-export function walkContent(node, fn) {
-  if (typeof node === 'string') {
-    return fn(node);
+// Walk an entry node searching for content leaves
+export function walkEntry(entry, fn) {
```
stasm: Why are you changing this? In its original form, `walkContent` was versatile and could walk over any object.  The previous version could be used to walk over the entire `entries` object at once.

src/runtime/gaiabuild/index.js
```diff
@@ -3,8 +3,6 @@
 import { fetch } from './io';
 import { View } from '../../bindings/gaiabuild/view';
 
-export { qps, walkContent } from '../../lib/pseudo';
```
stasm: Why are you removing this?  We need this in Gaia's build/multilocale.js.

src/lib/mocks.js
```diff
@@ -10,17 +9,12 @@ export const lang = {
   dir: 'ltr'
 };
 
-export function createEntriesFromAST(ast) {
-  let entries = Object.create(null);
-  for (let i = 0, node; (node = ast[i]); i++) {
-    entries[node.$i] = createEntry(node);
-  }
+export function createEntriesFromAST(entries) {
   return entries;
```
stasm: Maybe just remove this?

src/lib/pseudo.js
zbraniecki: the problem with the previous one was that:
```
{
  value: value
  attrs: attrs
  index: index
}
```

looks exactly the same as:

```
{
  one: 'one',
  two: 'two',
  other: 'other'
}
```

I need to separate walking entries from walking values.

src/bindings/gaiabuild/legacy/env.js
```diff
@@ -32,5 +32,5 @@ export class LegacyEnv extends Env {
 
 function createPseudoEntry(node, lang) {
   return createEntry(
-    walkContent(node, qps[lang.code].translate), lang);
+    walkEntry(node, qps[lang.code].translate), lang);
```
stasm: Let's use this opportunity and remove `lang` from args here.

src/lib/pseudo.js
stasm: We still want to recurse into `value` and `attrs`.  Maybe just filter out nodes with `type` defined which would exclude expressions and indexes?

tools/format.js
```diff
@@ -36,9 +36,15 @@ function singleline(formatted) {
                    .trim();
 }
 
+var lang = {
+  code:'en-US',
+  src: 'app',
+  dir: 'ltr'
+};
```
stasm: Can you import this from src/lib/mocks?

tools/perf/example.properties
```diff
@@ -586,7 +586,6 @@ build-id=Build Identifier
 update-channel=Update channel
 last-updated=Last Updated
 more-info=More Information
-gitInfo=Git commit info
```
stasm: Why remove this?

src/lib/env.js
zbraniecki: good catch. fixing in the next patch

src/lib/pseudo.js
zbraniecki: that's not enough. The problem is not about expression. The problem is that the shape of en entry is the same as the shape of the Hash. 

And I do recurse into value and attrs, but I use walkValue instead of walkEntry.

src/lib/env.js
stasm: That's OK.  Actually, it shouldn't report an error for inter-resource dupes. We do want to allow those.

src/runtime/gaiabuild/index.js
zbraniecki: thanks! fixing - I ignored the word "export" ;)

src/lib/mocks.js
zbraniecki: yup!

src/lib/format/l20n/parser.js
```diff
@@ -5,201 +5,134 @@ import { L10nError } from '../../errors';
...
-    let buf = this._source.slice(this._index + 1, opcharPos);
-
-    this._index = opcharPos + 1;
+    // skip '>'
+    ++this._index;
 
-    if (!this.simpleMode && buf.indexOf('\\') !== -1) {
-      buf = this.unescapeString(buf, opchar);
+    if (id in this.entries) {
+      throw this.error('Duplicate entry ID "' + id);
```
stasm: Can you use the same message for the error here as before?  Also, I think we need to somehow emit a `duplicaterror` here for the buildtime bindings to behave as expected.

tools/perf/example.properties
zbraniecki: because it's a duplicated entity and now the parser errors on it :) I'll update both example.properties and example.l20n to use current settings' resource in a separate patch.

src/lib/pseudo.js
stasm: OK, I see you've thought about this and I don't mind the change.  I'm just not sure if I understand why it's a problem that the shape is the same.

zbraniecki: consider:
```
<entity1 {
  value: "I'm Value 1",
  index: "I'm Value 2",
  attrs: "I'm Value 3"
}>
```

and how would it look like from the perspective of walkContent.

src/lib/format/l20n/parser.js
zbraniecki: I don't think I can use the same message because parser doesn't know which language it parses (and I think this separation is a good thing). What we could do is differentiate types of errors in the parser, and then in env when we catch duplicateentry error report it properly. What do you think?

src/lib/pseudo.js
stasm: Is that a valid entry though?

tools/perf/example.properties
stasm: nice!

src/lib/format/l20n/parser.js
stasm: But `emit` knows this :)  https://github.com/l20n/l20n.js/blob/c1a3c420052e757beb397a65fb99fd6b868fc9ac/src/lib/env.js#L39

src/lib/pseudo.js
zbraniecki: why not?

stasm: `attrs` should be an object.  I see your point about the `index`.  It can be a string and we shouldn't pseudolocalize it.

src/lib/format/l20n/parser.js
zbraniecki: omg. cool! But... how am I supposed to plug it in there?

stasm: `this.emit('duplicateerror', new L10nError(...))` and `this.emit('parseerror', new L10nError(...))`, depending on the error.

src/lib/env.js
zbraniecki: really? Dang. I was hoping we would report when a context has duplicate, inter-resource dupes. Because otherwise that may lead to very hard to discover bugs.

Also, I believe BTO reports that, right?

src/lib/format/l20n/parser.js
zbraniecki: Oh, so this?

```
      const error = this.error('Duplicate entry ID "' + id);
      if (this.emit) {
        throw this.emit('duplicateerror', error);
      } else {
        throw error;
      }
```

I'd like to work on the emitting code for runtime later to not have to check if emit is there, but I'd prefer not to add more to this patch.

src/lib/env.js
stasm: If we want to allow for mixing resources into 'views' freely, I think we should allow inter-resource dupes.  Also, this lets developers to define a resource which overrides entities of other resources.

zbraniecki: hmmm... the last argument is reasonable. I'm still scared of accidental overrides so let's make sure that we remember that resource order in a view is significant.
We should also probably fix BTO to not report them then as well.

src/lib/format/l20n/parser.js
stasm: You don't want to emit here, catch the error in `getResource` and re-emit as `parseerror` again.

zbraniecki: Ugh, I still fail to understand what you want me to do here.

So, you want me to throw an L10nError here and let the try...catch in getResource emit this as parsererror.

But in your original comment you are saying:
1) Emit the same error (which I cannot because the original error had lang.code in message)
2) somehow emit duplicateerror here, but you don't want me to emit.

I'm confused.

src/lib/env.js
stasm: Absolutely, it's a feature of the views system: the order is significant.  For BTO, I think it's okay to be a bit more lax with dupes in l20n.js, but detect all of them in (legacy) l10n.js.

src/lib/format/l20n/parser.js
stasm: No, I don't want to emit this as `parseerror`.  How about this:

```
error(type, msg) {
  // add context to msg
  const err = new L10nError(msg);
  this.emit(type, err);
  return err;
}
```

And then in other methods:

```
throw this.error('duplicateerror', 'Yo');
```

And don't emit in `getResource`.

stasm:
OK, everything looks great, just one final comment:  can we also have dupes detection for properties files?  We should disallow redefinitions of entity values (`foo=..`) and entity attributes (`foo.bar=..`).

PR #50, opened 2015-07-08
Support export ... from, babel 5.6 and don't modules between tasks

zbraniecki:


build/config/babel/bundler.js
```diff
@@ -3,8 +3,8 @@
 var Path = require('path');
 var fs = require('fs');
 
-var moduleIds = new Set();
-var modules = [];
+var moduleIds;
+var modules;
 
 /*
     const modules = new Map();
```
stasm: Update the comment?

build/config/babel/bundler.js
```diff
@@ -144,63 +142,73 @@ function addImports(babel, imports) {
...
+  } else {
+    var idents = [];
+    for (var j = 0; j < node.specifiers.length; j++) {
+      idents.push(babel.types.identifier(node.specifiers[j].local.name));
+    }
+    id = babel.types.objectPattern(idents);
+  }
+
+  var getModule = babel.types.callExpression(
+      babel.types.identifier('getModule'),
```
stasm: Please use 2 spaces for indentation.

build/config/babel/bundler.js
```diff
@@ -24,17 +24,14 @@ function getPreamble(babel) {
   var t = babel.types;
   var body = [];
   body.push(
-    t.expressionStatement(
-      t.literal('use strict')));
```
stasm: Why are you moving this from here?

build/config/babel/bundler.js
zbraniecki: but the code didn't change.

build/config/babel/bundler.js
zbraniecki: because I think it's better if it's at the top of the file (the whole program is strict) than it to be inside the closure (program is not strict, closure is strict).

stasm: If you put it at the top of the file, all scripts used by the app which included the bundled file will be run in strict mode.  The strict mode is turned on globally.  It's much safer to put `'use strict';` inside of the IIFE to turn it on just for the scope of the IIFE.

build/config/babel/bundler.js
stasm: I think I misread the babel code below.  What does `t.newExpression(t.identifier('Map'), []))` do?  In particular, the `[]` at the end?

build/config/babel/bundler.js
zbraniecki: oh really? ok. I'll update. TIL! Thanks :)

build/config/babel/bundler.js
zbraniecki: Not crash babel. Don't ask my why please. I just know that without the [] as a second argument babel crashed without whitelist. With it, it doesn't. :)

stasm: <3

Can we remove the whitelist from build/config/babel.index.js?

zbraniecki: No :( But we can make it empty.

PR #53, opened 2015-07-10
Serialize l20n to object

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1182406

### DIFF
--- a/src/bindings/gaiabuild/serialize.js
+++ b/src/bindings/gaiabuild/serialize.js
@@ -2,7 +2,20 @@
 
 import { L10nError } from '../../lib/errors';
 
-export function serializeEntries(lang, langEntries, sourceEntries) {
+export function serializeContext(ctx, lang) {
+  const cache = ctx._env._resCache;
+  return ctx._resIds.reduce(([errorsSeq, entriesSeq], cur) => {
+    const sourceRes = cache[cur + 'en-USapp'];
+    const langRes = cache[cur + lang.code + lang.src];
+    const [errors, entries] = serializeEntries(
+      lang,
+      langRes instanceof L10nError ? {} : langRes,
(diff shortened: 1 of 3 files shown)


stasm:
@zbraniecki r?

src/bindings/gaiabuild/serialize.js
```diff
@@ -30,6 +43,14 @@ export function serializeEntries(lang, langEntries, sourceEntries) {
   return [errors, entries];
 }
 
+function extend(into, from) {
+  for (let key in from) {
+    // overwrite existing keys for reduceRight
+    into[key] = from[key];
```
zbraniecki: from is a keyword in ES6. I doubt it'll be a syntax error, but maybe use a different variable name?

src/bindings/gaiabuild/serialize.js
```diff
@@ -2,7 +2,20 @@
...
+export function serializeContext(ctx, lang) {
+  const cache = ctx._env._resCache;
+  return ctx._resIds.reduceRight(([errorsSeq, entriesSeq], cur) => {
+    const sourceRes = cache[cur + 'en-USapp'];
+    const langRes = cache[cur + lang.code + lang.src];
+    const [errors, entries] = serializeEntries(
+      lang,
+      langRes instanceof L10nError ? {} : langRes,
+      sourceRes instanceof L10nError ? {} : sourceRes);
+    return [errorsSeq.concat(errors), extend(entriesSeq, entries)];
```
zbraniecki: Is there a way to use Object.assign here instead of custom extend?

stasm: Ah, perfect, just what I needed! Thanks.

PR #54, opened 2015-07-10
Support the legacy AST in the resolver

stasm:
@zbraniecki, I think we missed one thing when we landed the runtime parser:  the resolver needs to support both modes :(

This PR is based on https://github.com/l20n/l20n.js/commit/56bdee68c0eed0c9101db0bcede1596ca07032be#diff-36c4e7c2ec7b3cb60316760123cc81e9.

### DIFF
--- a/src/bindings/gaiabuild/legacy/env.js
+++ b/src/bindings/gaiabuild/legacy/env.js
@@ -2,62 +2,45 @@
 
 import { L10nError } from '../../../lib/errors';
 import { Env, amendError } from '../../../lib/env';
+import { LegacyContext } from './context';
 import { createEntry } from './resolver';
 import PropertiesParser from './parser';
 import { walkContent } from './pseudo';
 import { qps } from '../../../lib/pseudo';
-import {
-  emit, addEventListener, removeEventListener
-} from '../../../lib/events';
 
-// XXX babel transpiles class inheritance to code which modifies prototypes 
-// which triggers warnings in Gecko;  we redefine LegacyEnv from scratch to 
-// avoid warnings.
-export class LegacyEnv {
-  constructor(defaultLang, fetch) {
-    // XXX babel doesn't allow calling the constructor as a regular function 
-    // so we need to copy the whole constructor from Env
-    this.defaultLang = defaultLang;
-    this.fetch = fetch;
-
-    this._resCache = Object.create(null);
-
-    const listeners = {};
-    this.emit = emit.bind(this, listeners);
-    this.addEventListener = addEventListener.bind(this, listeners);
-    this.removeEventListener = removeEventListener.bind(this, listeners);
-  }
-
-  createContext(resIds) {
-    return Env.prototype.createContext.call(this, resIds);
-  }
-
-  _parse(syntax, lang, data) {
-    const emit = (type, err) => this.emit(type, amendError(lang, err));
-    return PropertiesParser.parse.call(PropertiesParser, emit, data);
-  }
-
-  _create(lang, ast) {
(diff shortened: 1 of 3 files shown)


src/lib/resolver.js
```diff
@@ -108,7 +108,8 @@ function interpolate(locals, ctx, lang, args, arr) {
     if (typeof cur === 'string') {
       return [localsSeq, valueSeq + cur];
     } else {
-      const [, value] = subPlaceable(locals, ctx, lang, args, cur.name);
+      const [, value] = subPlaceable(
+        locals, ctx, lang, args, cur.name || cur.v);
```
zbraniecki: why is that cur.name or cur.v? Is that because Identifier has `v` in old AST and `name` in the new one?

src/lib/resolver.js
```diff
@@ -130,8 +131,9 @@ function resolveSelector(ctx, lang, args, expr, index) {
     return selector;
   }
 
-  const argValue = index[0].args ?
-    resolveIdentifier(ctx, lang, args, index[0].args[0].name)[1] : undefined;
+  const argName = index[0].args ? index[0].args[0].name : index[1];
+  const argValue = argName ?
+    resolveIdentifier(ctx, lang, args, argName)[1] : undefined;
```
zbraniecki: Would it make more sense to actually fork resolver as well in that scenario? I'm worried about trying to resolve two different AST's in one resolver and introducing legacy fallbacks in our new resolver.

stasm: Wait, there's something I don't understand.  I did for the resolver into src/bindings/gaiabuild/legacy/resolver.  I'm thinking about this with a fresh mind and I don't see why this PR should be needed.  It does solve https://bugzil.la/1182568 though.  I need to investigate more.

stasm: OK, I see now: the forked resolver only has `createEntry` and `createAttribute`.  `format` comes from the regular resolver.

stasm: Let's fork it;  we won't be adding new features to the legacy resolver so it should be easy to maintain it.  OTOH, like you said, supporting both ASTs in one is going to get harder and harder

stasm:
@zbraniecki New approach, can you take a look?

src/bindings/gaiabuild/legacy/env.js
```diff
@@ -2,62 +2,45 @@
...
-  emit, addEventListener, removeEventListener
-} from '../../../lib/events';
 
-// XXX babel transpiles class inheritance to code which modifies prototypes 
-// which triggers warnings in Gecko;  we redefine LegacyEnv from scratch to 
-// avoid warnings.
-export class LegacyEnv {
-  constructor(defaultLang, fetch) {
-    // XXX babel doesn't allow calling the constructor as a regular function 
-    // so we need to copy the whole constructor from Env
```
stasm: I found a workaround for this:  using `LegacyEnv.prototype = Object.create(Env.prototype);`, `this` has `Env` in its prototypes chain, so babel's `classCallCheck` doesn't complain anymore.

zbraniecki: is it implementation detail in babel and may break in future babel versions?

stasm: Isn't all of Babel an 'implementation detail'? :)

stasm: On a more serious note, it might break in the future. I'm not sure why they enforce this class constructor check. I might file an issue with babel to remove it.

PR #55, opened 2015-07-15
Move fetch out of format* methods in Context

stasm:
We can optimize for the most common scenario by prefetching the resources for the first languages once instead of running the same logic on every `formatValue` or `formatEntity` call.


stasm:
@zbraniecki r?

PR #56, opened 2015-07-15
Remove formatValue

stasm:
This is based on #55.

The distinction between `formatValue` and `formatEntity` makes the `Context` code more complex and creates an impression that attributes are second-class.  OTOH, the HTML bindings only use `formatEntity` anyways.  It's trivial to shim the bahavior of `formatValue` if needed.

Before:

```javascript
document.l10n.formatValue('foo').then(value => alert(value));
```
After:

```javascript
document.l10n.format('foo').then(({value}) => alert(value));
```

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -111,7 +111,7 @@ function getElementTranslation(view, langs, elem) {
   const l10n = getAttributes(elem);
 
   return l10n.id ?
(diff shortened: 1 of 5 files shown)


stasm:
In Gaia, the Notification Helper and the Gallery app use `formatValue` but it's the one defined on `navigator.mozL10n` so this code will need to be updated anyways.

stasm:
@zbraniecki r?

PR #57, opened 2015-07-15
Add position to JunkEntry.

Pike:
For aisle, I need position info in the JunkEntries.

Note, I also fixed a bug compared to my previous version, which was giving the wrong start position for the JunkEntry

### DIFF
--- a/src/lib/format/l20n/ast/parser.js
+++ b/src/lib/format/l20n/ast/parser.js
@@ -483,7 +483,11 @@ export default {
 
     this._index = nextEntry;
 
-    return new AST.JunkEntry(
+    const junk = new AST.JunkEntry(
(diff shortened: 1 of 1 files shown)


src/lib/format/l20n/ast/parser.js
```diff
@@ -483,7 +483,11 @@ export default {
 
     this._index = nextEntry;
 
-    return new AST.JunkEntry(
+    const junk = new AST.JunkEntry(
       this._source.slice(this._curEntryStart, nextEntry));
+    if (this._config.pos) {
+      junk._pos = {start: this._curEntryStart, end: nextEntry};
```
stasm: Do we need this to be prefixed with an underscore?  Maybe just `junk.pos`?

PR #58, opened 2015-07-16
bug 1184498, error recovery by creating minimal and isolated JunkEntries

Pike:
When finding parser errors, make the JunkEntry as small as possible,
and continue parsing.
If there are adjancent JunkEntries, glue them together.
After parsing, getEntry on just the JunkEntry to create error messages.

Requesting feedback for now, we'd need to port this to the entries parser, too.


stasm:
Closing this as we now have a new syntax and a new parser.  Thanks, @Pike!

PR #59, opened 2015-07-20
Bug 1185678 - Create Parse Context object for L20n Parsers

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1185678

### DIFF
--- a/src/lib/format/l20n/ast/parser.js
+++ b/src/lib/format/l20n/ast/parser.js
@@ -488,12 +488,11 @@ class ParseContext {
 
     const junk = new AST.JunkEntry(
       this._source.slice(this._curEntryStart, nextEntry));
-    if (this._config.pos) {
-      junk._pos = {start: this._curEntryStart, end: nextEntry};
-    }
+
+    this.setPosition(junk, this._curEntryStart, nextEntry);
     return junk;
   }
-};
+}
 
 export default {
   parseResource: function(string, pos = false) {
--- a/src/lib/format/l20n/ast/parser.js
+++ b/src/lib/format/l20n/ast/parser.js
@@ -5,25 +5,28 @@ import { L10nError } from '../../../errors';
 
 const MAX_PLACEABLES = 100;
 
-export default {
-  parse: function(env, string, pos = false) {
+
+class ParseContext {
+  constructor(string, pos) {
+    this._config = {
+      pos: pos
+    };
     this._source = string;
     this._index = 0;
     this._length = string.length;
     this._curEntryStart = 0;
-    this._config = {
-      pos: pos
-    };
+  } 
 
-    if (pos !== true) {
-      AST.Node.prototype.setPosition = function() {};
+  setPosition(node, start, end) {
+    if (!this._config.pos) {
+      return;
     }
-    return this.getResource(pos);
-  },
+    node._pos = {start, end};
+  }
 
-  getResource: function() {
+  getResource() {
     let resource = new AST.Resource();
-    resource.setPosition(0, this._length);
+    this.setPosition(resource, 0, this._length);
     resource._errors = [];
 
     this.getWS();
@@ -44,9 +47,9 @@ export default {
     }
 
     return resource;
-  },
+  }
 
-  getEntry: function() {
+  getEntry() {
     this._curEntryStart = this._index;
 
     if (this._source[this._index] === '<') {
@@ -64,9 +67,9 @@ export default {
     }
 
     throw this.error('Invalid entry');
-  },
+  }
 
-  getEntity: function(id, index) {
+  getEntity(id, index) {
     if (!this.getRequiredWS()) {
       throw this.error('Expected white space');
     }
@@ -94,11 +97,11 @@ export default {
     ++this._index;
 
     const entity = new AST.Entity(id, value, index, attrs);
-    entity.setPosition(this._curEntryStart, this._index);
+    this.setPosition(entity, this._curEntryStart, this._index);
     return entity;
-  },
+  }
 
-  getValue: function(ch = this._source[this._index], optional = false) {
+  getValue(ch = this._source[this._index], optional = false) {
     switch (ch) {
       case '\'':
       case '"':
@@ -111,17 +114,17 @@ export default {
       throw this.error('Unknown value type');
     }
     return null;
-  },
+  }
 
-  getWS: function() {
+  getWS() {
     let cc = this._source.charCodeAt(this._index);
     // space, \n, \t, \r
     while (cc === 32 || cc === 10 || cc === 9 || cc === 13) {
       cc = this._source.charCodeAt(++this._index);
     }
-  },
+  }
 
-  getRequiredWS: function() {
+  getRequiredWS() {
     const pos = this._index;
     let cc = this._source.charCodeAt(pos);
     // space, \n, \t, \r
@@ -129,9 +132,9 @@ export default {
       cc = this._source.charCodeAt(++this._index);
     }
     return this._index !== pos;
-  },
+  }
 
-  getIdentifier: function() {
+  getIdentifier() {
     const start = this._index;
     let cc = this._source.charCodeAt(this._index);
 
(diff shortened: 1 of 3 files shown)


src/lib/format/l20n/ast/parser.js
```diff
@@ -7,18 +7,20 @@ const MAX_PLACEABLES = 100;
 
 export default {
   parse: function(env, string, pos = false) {
```
stasm: Can we make 'parse' the only exported symbol?

    export function parse(emit, string, pos = false) { ...}

This will allow us to do:

    import { parse } from 'format/l20n/ast/parser';

src/lib/format/l20n/ast/parser.js
```diff
@@ -7,18 +7,20 @@ const MAX_PLACEABLES = 100;
...
-    return this.getResource(pos);
+    AST.Node._config.pos = pos;
+
+    const parseContext = this.getNewParseContext(env, string, pos);
+
+    return parseContext.getResource();
+  },
+
+  getNewParseContext: function(env, string, pos = false) {
+    const parseContext = Object.create(this);
```
stasm: Maybe all of the internal parsing functions should simply be methods on `ParseContext.prototype`?

stasm:
@Pike f?

src/lib/format/l20n/ast/parser.js
```diff
@@ -485,7 +487,7 @@ export default {
 
     const junk = new AST.JunkEntry(
       this._source.slice(this._curEntryStart, nextEntry));
-    if (this._config.pos) {
+    if (AST.Node._config.pos) {
```
Pike: I think I prefer the pos state to be on the parser state, and not on AST.Node.

Note, we'd expose that config on each node we return from the parse cycle, right?

Pike:
Not a whole to add, besides the nit on state exposed on the returned AST.Node children. Which would weirdly jump if you parse with and without pos?

src/lib/format/l20n/ast/parser.js
zbraniecki: we potentially may want to have sth like parseString, or parseEntity.

src/lib/format/l20n/ast/parser.js
```diff
@@ -7,18 +7,20 @@ const MAX_PLACEABLES = 100;
...
-      pos: pos
-    };
-
-    if (pos !== true) {
-      AST.Node.prototype.setPosition = function() {};
-    }
-    return this.getResource(pos);
+    AST.Node._config.pos = pos;
+
+    const parseContext = this.getNewParseContext(env, string, pos);
```
zbraniecki: For ParseContext to be newable it would have to be a function, right now it's an object.

src/lib/format/l20n/ast/parser.js
zbraniecki: Are you suggesting switching ParseContext to be a class?

src/lib/format/l20n/ast/parser.js
zbraniecki: I agree. In the updated version I moved it back and actually turned `setPosition` to be a method on the parser itself that turns into a no-op if `_config.pos` is false.

src/lib/format/l20n/ast/parser.js
stasm: Can we choose an unambiguous and future-proof name and make it the only exported symbol right now?

src/lib/format/l20n/ast/parser.js
stasm: Yes and no, depending on what you mean by _class_ given that this is JavaScript :)  With `const parseContext = Object.create(this);` you're essentially doing exactly what the `new` keyword does: creating a new object with a prototype set to another object.

By doing this in a more idiomatic way, the code might end up being more maintainable and also we might benefit from Gecko optimizations targeted at object instantiation.

So yeah, maybe the easiest way of achieving the same thing would be simply to have something like:

```javascript
export function parseString(str) {
  ...
  const pctx = new ParserContext(...);
  ...
}

function ParserContext(...) {
   this.index = 0;
   ...
}

ParserContext.prototype.getResource = function(...) { ... }
etc.
```

(or use es6 classes, it's the same thing.)

src/lib/format/l20n/ast/parser.js
stasm: I'm not sure if I understand this comment.

src/lib/format/l20n/ast/parser.js
zbraniecki: yeah, is parseResource ok? (trying to name it similar to parseProgram from JS and parseStylesheet from CSS)

src/lib/format/l20n/ast/parser.js
zbraniecki: Currently, there's no `ParseContext` that could be used in a snippet that you suggested. And if I name the default exported object `ParseContext` I still cannot do `new ParseContext()` because it is an object, not a function:

```
const ParseContext = {};
const parseContext = new ParseContext()
```

So maybe I don't understand your comment.

src/lib/format/l20n/ast/parser.js
zbraniecki: ok. So I think that you are suggesting to turn ParserContext to a class :) I'll work on that!

src/lib/format/l20n/ast/parser.js
stasm: `ParseContext` would come from https://github.com/l20n/l20n.js/pull/59#discussion_r36062939 :)  Let's discuss there.

src/lib/format/l20n/ast/parser.js
stasm: I like it!

zbraniecki:
@stasm f?

src/lib/format/l20n/ast/parser.js
```diff
@@ -491,3 +498,10 @@ export default {
     return junk;
   }
 };
+
+export default {
+  parse: function(env, string, pos = false) {
```
stasm: Should this also be called `parseResource`?

stasm: Also, `env` is not used in the code.  And what happened to `emit`?

stasm: Ah, sorry, this is the AST parser.  No `emit`.  Let's remove `env` too?

src/lib/format/l20n/ast/parser.js
```diff
@@ -5,25 +5,32 @@ import { L10nError } from '../../../errors';
...
-  parse: function(env, string, pos = false) {
-    this._source = string;
-    this._index = 0;
-    this._length = string.length;
-    this._curEntryStart = 0;
+
+class ParseContext {
+  constructor(env, pos) {
     this._config = {
       pos: pos
```
stasm: Would it make sense to flatten this into `this._configPos` since there's only one config item right now?

zbraniecki: since it's not a runtime, I think I prefer to keep it as a collection and let it grow. But that's an API preference, not based on any strong reasons.

@Pike : any preference here?

src/lib/format/l20n/ast/parser.js
```diff
@@ -5,25 +5,32 @@ import { L10nError } from '../../../errors';
...
-    return this.getResource(pos);
-  },
+    node._pos = {start, end};
+  }
+
+  parseResource(string) {
+    this._source = string;
+    this._index = 0;
+    this._length = string.length;
+    this._curEntryStart = 0;
```
stasm: I think these should be set in the constructor?  This way it will be possible to create a new `ParseContext` and call _any_ method.  This is what #62 might want to use, for instance, with `getEntry`.

tools/lib/index.js
```diff
@@ -9,7 +9,7 @@ exports.parse = function(fileformat, output, text) {
   var module = fileformat === 'properties' ?
     '../../src/lib/format/properties/parser' :
     '../../src/lib/format/' + fileformat + '/' + output + '/parser';
-  return require(module).parse(null, text);
+  return require(module).parseResource(text);
```
stasm: Support the entries parser here, too, please.

stasm:
Has this landed in https://bugzilla.mozilla.org/show_bug.cgi?id=1185678 ?

zbraniecki:
yup

PR #60, opened 2015-07-23
add grunt-cli so global install is not required

jedireza:
Without instructions, I cloned this repo, ran `$ npm install` and then `$ npm test`, which failed because `grunt` wasn't installed globally.

By adding `grunt-cli` to the `devDependencies`, a local `grunt` binary is created and doesn't require a global installation of `grunt` when using `grunt` via npm scripts.

This change also makes it unnecessary to install `grunt-cli` globally on Travis. Which I also removed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
   - "0.12"
-before_script:
-  - npm install -g grunt-cli
 notifications:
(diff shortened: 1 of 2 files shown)


zbraniecki:
love it!

stasm:
Sweet, thanks!

PR #63, opened 2015-07-31
Remove the Babel Polyfill requirement

stasm:
The `for..of` syntax requires iterators which are only supported by Babel with
a special polyfill which emulates a full ES6 environment.  The polyfill is
quite hefty weighing at 25KB.  By replacing `for..of` with a different method
of iterating over an array we can remove this requirement and make l20n.js
more lightweight in the webcompat version.

http://babeljs.io/docs/usage/polyfill/

Additionally, I suggest that we turn on Babel's 'loose' transpilation.  From
their page:

    Loose mode enables certain transformers to generate cleaner output that
    lacks specific ES6 edgecases. These edgecases are either unlikely to
    appear in your code or the inclusion of them introduces significant
    overhead.

    As a result of loose mode code will execute faster and be much more
    readable and comparable to the original but will deviate from the spec
    in slight ways.

https://babeljs.io/docs/advanced/loose/

Lastly, Babel also includes some optional transformers which make the code
cleaner.  Let's turn them on as well.

http://babeljs.io/docs/advanced/transformers/#optional

### DIFF
--- a/src/bindings/html/head.js
+++ b/src/bindings/html/head.js
@@ -16,7 +16,7 @@ export function getMeta(head) {
     'meta[name="availableLanguages"],' +
     'meta[name="defaultLanguage"],' +
     'meta[name="appVersion"]');
-  for (let el of els) {
+  Array.prototype.forEach.call(els, el => {
     const name = el.getAttribute('name');
     const content = el.getAttribute('content').trim();
     switch (name) {
@@ -34,7 +34,8 @@ export function getMeta(head) {
       case 'appVersion':
         appVersion = content;
     }
-  }
+  });
+
   return {
     defaultLang,
     availableLangs,
--- a/src/bindings/html/head.js
+++ b/src/bindings/html/head.js
@@ -12,13 +12,17 @@ export function getMeta(head) {
   let appVersion = null;
 
   // XXX take last found instead of first?
-  const els = head.querySelectorAll(
+  const metas = head.querySelectorAll(
     'meta[name="availableLanguages"],' +
     'meta[name="defaultLanguage"],' +
     'meta[name="appVersion"]');
-  Array.prototype.forEach.call(els, el => {
-    const name = el.getAttribute('name');
-    const content = el.getAttribute('content').trim();
+
+  // Polyfill NodeList.prototype[Symbol.iterator] for Chrome.
+  // See https://code.google.com/p/chromium/issues/detail?id=401699
+  metas[Symbol.iterator] = Array.prototype[Symbol.iterator];
+  for (let meta of metas) {
+    const name = meta.getAttribute('name');
+    const content = meta.getAttribute('content').trim();
     switch (name) {
       case 'availableLanguages':
         availableLangs = getLangRevisionMap(
@@ -34,7 +38,7 @@ export function getMeta(head) {
       case 'appVersion':
         appVersion = content;
     }
-  });
+  }
 
   return {
     defaultLang,
--- a/src/bindings/html/head.js
+++ b/src/bindings/html/head.js
@@ -1,6 +1,10 @@
 'use strict';
 
-import './polyfill';
+// Polyfill NodeList.prototype[Symbol.iterator] for Chrome.
+// See https://code.google.com/p/chromium/issues/detail?id=401699
+if (!NodeList.prototype[Symbol.iterator]) {
+  NodeList.prototype[Symbol.iterator] = Array.prototype[Symbol.iterator];
+}
 
 export function getResourceLinks(head) {
   return Array.prototype.map.call(
--- a/src/bindings/html/head.js
+++ b/src/bindings/html/head.js
@@ -1,5 +1,7 @@
 'use strict';
 
+import './polyfill';
+
 export function getResourceLinks(head) {
   return Array.prototype.map.call(
     head.querySelectorAll('link[rel="localization"]'),
@@ -16,10 +18,6 @@ export function getMeta(head) {
     'meta[name="availableLanguages"],' +
     'meta[name="defaultLanguage"],' +
     'meta[name="appVersion"]');
-
-  // Polyfill NodeList.prototype[Symbol.iterator] for Chrome.
-  // See https://code.google.com/p/chromium/issues/detail?id=401699
-  metas[Symbol.iterator] = Array.prototype[Symbol.iterator];
   for (let meta of metas) {
     const name = meta.getAttribute('name');
     const content = meta.getAttribute('content').trim();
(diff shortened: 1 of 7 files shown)


stasm:
The _loose_ mode makes our files smaller, sometimes by more than 10 kB:

`webcompat/l20n.js`  71.7 kB → 62.9 kB
`gaiabuild/l20n.js`  89.9 kB → 78.2 kB
`tooling/l20n.js` 104 kB → 91.4 kB
`aisle/l20n.js` 23 kB → 20.8 kB
`node/l20n.js` 47 kB → 42.7 kB

stasm:
@zbraniecki ?

zbraniecki:
So, that sounds like you want us to not use `for..of` syntax at all... for how long?

Can we do something smart instead and create babel plugin that transforms `for..of` for where we need compatibility layer?

stasm:
I think a custom plugin is a lot of work for not much benefits.  `for..of` has its special semantics related to iteration.  I'd prefer to not have two different implementations with two different semantics.  Babel requires the polyfill to make sure the semantics are as close to the spec as possible.

I'm okay with using `for..of` in gaiabuild bindings for instance or tooling, where we don't need to worry about the size of the package.

stasm:
What I mean is:  if you're okay with us transpiling `for..of` to something else which has different semantics, then maybe we don't care about `for..of`'s semantics after all.  In which case let's just use something else :)

zbraniecki:
Actually, my thinking goes like this: I want to work in es6, using es6 semantics. `for..of` is going to be increasingly important part of that because I want to use Set/Map in place of Array/Object where appropriate and unfortunately es6 syntax makes almost everything with Set/Map harder *except* of iterating and for that using `for..of`.

From target audience that I see for l20n.js there's a bunch of targets that are gaining es6 compatibility really fast (so, in half a year I can imagine them being on par) like Gecko, Webkit, Blink, Node, Edge. And one, IE11, which will never move forward if I'm correct.

Blocking ourselves from using `for..of` semantics in order to maintain that backward compatibility long term sounds like a bad strategic move, so I'm looking for alternatives.

I'm wondering if it would be possible to write a babel plugin that would turn `for..of` for targets that have IE11 support goal into something that works for that target yet is significantly simpler than the polyfill that Babel offers (because we know our uses of `for..of` we can exclude edge cases).

Does it make sense?

stasm:
I see where you're coming from, I just don't think I agree on the "targets that have IE11 support goal" part.  I think we have two targets: 1) Gaia and 2) all of the Web, which includes IE.  I'd like to monitor the adoption of Edge and I wish all the best, and if it ends up replacing most of IE1* that's great.  Let's revisit this in 6 months or a year, and let's keep in mind the 10-15% of users which we can support _right now_ with this simple change.

I'm not sure I like the babel plugin idea.  It's more work and also it introduces two implementations into our codebase which might make it harder to maintain.  Sets and Maps can use their native `forEach` methods, too, so I think replacing `for..of` with `forEach` makes sense in our current situation.  Would you agree?

zbraniecki:
I think I'm less worried about the patch than about setting a precedence and establishing a culture of writing l20n.js tied to what IE11 supports. 
I know you have a goal of having html and gaia use one binding, and I agree with you, but maybe for the time being we could fork those two in order to achieve two goals:
 * keep using es6 syntax in gaia bindings and backport for html bindings
 * allow us to use gaia/gecko APIs in gaia bindings (langpack API doesn't make sense in html bindings for example).

I even imagine that once we start implementing the APIs for Locale/LanguageNegotiation/Resource loading and preventing layout that are Gecko/Gaia specific that separation will make even more sense.

On the other hand if we'll get all browsers that we want to support to support `for..of` and the API will not be landing, we could merge the two bindings.

tl;dr - I'm ok with the patch. I'm more worried about what it means for how we'll be writing code over next months.

zbraniecki:
Also, is it only `for..of`? What about destructing assignments? default values? shorthand objects? rest params, template strings, proxies, symbols etc?

stasm:
AFAICT, out of our current es6 usage, the polyfill is only required for `for..of`.  Other features have more straightforward transpilations.

zbraniecki:
ok. that sounds less scary. I think I'm ok with sacrificing `for..of` then. Can we define the conditions for when are we going to lift the ban on `for..of`? :)

stasm:
I did some more research and it turns out that the reason why Chrome needed the polyfill was that it doesn't define `Symbol.iterator` on `NodeLists`, so the following is not possible:

```javascript
for (let elem of document.querySelectorAll(...)) { ... }
```

Otherwise, `for..of` works well.

They're working on adding iteration to `NodeLists` in https://code.google.com/p/chromium/issues/detail?id=401699.  In the meantime, we can simply convert them to arrays or do what https://jakearchibald.com/2014/iterators-gonna-iterate/ suggests: 

```javascript
NodeList.prototype[Symbol.iterator] = Array.prototype[Symbol.iterator];
```

So, to sum up:

 - Firefox supports everything well,
 - Chrome doesn't define `NodeList.prototype[Symbol.iterator]` but the workarounds are easy,
 - Edge will be great,
 - IE11 is hopeless and doesn't support almost anything.  Let's work in #64 on our strategy for IE11, and in this pr let's just remove the code which requires the polyfill for Chrome.

src/bindings/html/head.js
```diff
@@ -12,13 +12,17 @@ export function getMeta(head) {
...
     'meta[name="availableLanguages"],' +
     'meta[name="defaultLanguage"],' +
     'meta[name="appVersion"]');
-  for (let el of els) {
-    const name = el.getAttribute('name');
-    const content = el.getAttribute('content').trim();
+
+  // Polyfill NodeList.prototype[Symbol.iterator] for Chrome.
+  // See https://code.google.com/p/chromium/issues/detail?id=401699
+  metas[Symbol.iterator] = Array.prototype[Symbol.iterator];
```
zbraniecki: should we do this only if we it doesn't exist?

stasm: Sure.

zbraniecki: Or, can we do this in one place instead:
```
if (!NodeList.prototype[Symbol.iterator]) {
  NodeList.prototype[Symbol.iterator] = Array.prototype[Symbol.iterator];
}
``` 

it's basically a polyfill.

stasm: It is, yes.  I'm not sure where to put it.  We need it in html/head and we might start using `for..of` in html/dom (but we don't right now).  Maybe add a new module, html/polyfill, which can be required by any other code?

src/bindings/html/head.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+import './polyfill';
```
stasm: This gets transpiled to `const {} = getModule('bindings/html/polyfill');`.  @zbraniecki, can you help me fix this in our babel bundler?

zbraniecki:
r+

PR #64, opened 2015-07-31
Support IE 11

stasm:
I'd like to get us to a point where l20n.js works well in IE 11.  So far I discovered the following list of incompatibilities:

 - `navigator.languages` is undefined, need to use `[navigator.language]`,
 - no `WeakSet` but `Set` is available (#44 would help),
 - no `Promise`, needs a polyfill,
 - `CustomEvent` is not a constructor, need to use `document.createEvent`,
 - …?

I'll keep adding here if I find more. 

(The PR is a WIP.)

### DIFF
--- a/src/bindings/html/service.js
+++ b/src/bindings/html/service.js
@@ -22,14 +22,15 @@ export class Service {
       err => console.warn(err));
   }
 
-  requestLanguages(requestedLangs = navigator.languages) {
+  requestLanguages(requestedLangs) {
     return changeLanguages.call(
       this, getAdditionalLanguages(), requestedLangs);
   }
 
   handleEvent(evt) {
     return changeLanguages.call(
-      this, evt.detail || getAdditionalLanguages(), navigator.languages);
+      this, evt.detail || getAdditionalLanguages(),
+      navigator.languages || [navigator.language]);
   }
 }
 
--- a/src/runtime/web/index.js
+++ b/src/runtime/web/index.js
@@ -26,7 +26,7 @@ function init() {
   const service = new Service(fetch);
   window.addEventListener('languagechange', service);
   document.addEventListener('additionallanguageschange', service);
-  document.l10n.languages = navigator.languages;
+  document.l10n.languages = navigator.languages || [navigator.language];
 }
 
 whenInteractive(init);


zbraniecki:
I'm ok with that, but since IE11 is going to be obsolete by Edge soon, I'd prefer us not to block ourselves on using ES6 in order to support that.

Can we add polyfills for IE11 into our web target output and not affect our gaia target?

stasm:
It's not that we'll just add polyfills for IE11.  Any web developer will need to add them always, which effectively increases the size of the l20n.js runtime for all.

zbraniecki:
That's true, but for majority of browsers it's a very temporary state. I believe that a developer who's getting into using l20n.js in his project right now can estimate that in a matter of 3-6 months most of the polyfills won't be needed as long as he's targeting evergreens.

I'd prefer not to give up on WeakSets and Promises from your list in order to achieve compatibility with IE11. I'd prefer to use polyfills for them.

Alternatively we could fork the files that use those features and maintain two versions - one for no-polyfill (gaia) version, one for the web. What do you think?

SunboX:
Keep in mind, that there are a lot of Windows Phone 8.1 devices that will not get an update and even the new Windows Phone 10 version is not Edge. :smirk:

stasm:
In https://bugzilla.mozilla.org/show_bug.cgi?id=939994 we want to explicitly include IE11 as one of the supported browsers.

SunboX:
depends on: https://bugzilla.mozilla.org/show_bug.cgi?id=1217402

SunboX:
Btw, for Promise polyfilling I'm using this one on Windows Phone:
https://github.com/jakearchibald/es6-promise

zbraniecki:
I created a branch  - https://github.com/zbraniecki/l20n.js/tree/ie11 and got the example webcompat.html to work with ie11.

I'm generally leaning against letting IE11 affect our work, but I just found WeakSet polyfill, and I begin to think that maybe we may be able to polyfill the shit out of it.

If we can polyfill Promise, WeakSet, Template Tag etc. and the only think is how babel transpiles `for .. of` then maybe we can fix it in babel?

stasm:
I'm closing this.  We don't support IE11 on master.

PR #65, opened 2015-08-01
[discuss/wip] node and node + browserify/webpack compatibility

jedireza:
As you're already aware :wink: I think it's important to be friendly with server side node and also browser apps built with `browserify`/`webpack`.

I haven't tested the bundle in the browser yet, but I wanted to get this diff up so we can discuss some of these ideas.

- - - - -

ES6 modules (`import`/`export`) are not in node yet. So instead of linking the `main` directive in the `package.json` to the `./src/runtime/node` scripts, we could create a node distribution that get's put in the `./dist/node` folder. This can be done by using a [`prepublish` script](https://docs.npmjs.com/misc/scripts) that runs `grunt webpack` to ensure the `./dist/node` is created before publishing to npm. Doing this should also help make us more ES5 compatible.

You'll also notice a new `browser` directive in the `package.json` file. This allows tools like `browserify`/`webpack` to use a browser compatible version that uses the `./src/runtime/web/io` (in order to use XHR) instead of the the `./src/runtime/node/io` (which uses the `fs` module). This also helps us be more npm friendly by not forcing browser app makers to use `bower` if they'd prefer to use `npm` only.

- - - - -

I added some basic entity fixtures and a test script (more like a raw demo) to demonstrate getting this working with node. You can run it like this:

```bash
$ grunt webpack
$ npm run test-node

# > l20n@4.0.0-pre test-node /projects/l20n.js
# > node ./tests/node

# SINGLE VALUE:  { value: '¿Qué puedo pedir?', attrs: null }
# MULI VALUES [ { value: '¿Qué puedo pedir?', attrs: null },
#   { value: 'Open &lt;App&gt;', attrs: null },
#   { value: '&lt;Número&gt; De Llamada', attrs: null },
#   { value: 'notFound', attrs: null } ]
```


tests/node/index.js
```diff
@@ -0,0 +1,32 @@
...
+  {code: 'en-US'},
+  {code: 'fr'}
+];
+
+ctx.fetch(langs).then(init);
+
+function init () {
+  ctx.resolve(langs, 'whatCanIAsk').then(function (value) {
+    console.log('SINGLE VALUE: ', value);
+  });
```
jedireza: I'd prefer we didn't have to pass `langs` in each time we wanted to resolve an entity. We've already defined a default locale when creating the environment and even after that we fetch the `langs` via the `context`. I would rather the `resolve` method on `context` not have the first `langs` argument or let it be optional.

Maybe we could assign the `langs` to the context once and not need to pass them in every time.

tests/node/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+  Promise.all([
+    ctx.resolve(langs, 'whatCanIAsk'),
+    ctx.resolve(langs, 'openApp'),
+    ctx.resolve(langs, 'callNumber'),
+    ctx.resolve(langs, 'notFound')
+  ]).then(function (values) {
+    console.log('MULI VALUES', values);
+  });
```
jedireza: The `langs` comment I made above seems more obvious when we want multiple entity values.

I spoke with @zbraniecki a couple days ago about being able to pass an array of entities instead of using `Promise.all`. He liked the idea but also mentioned others may not.

Another thing I don't like much about using `Promise.all` is that the results come back in an array. Also, when getting multiple entities resolved, I'd prefer to have the results as a dictionary instead, since I'd probably end up producing one myself for easier digestion in my app.

Maybe multiple entity results could look something like this.

```js
{
  whatCanIAsk: { value: '¿Qué puedo pedir?', attrs: null },
  openApp: { value: 'Open &lt;App&gt;', attrs: null },
  callNumber: { value: '&lt;Número&gt; De Llamada', attrs: null },
  notFound { value: 'notFound', attrs: null }
}
```

To speak further to translating multiple entities... I could imagine producing a huge array of `ctx.resolve(langs, 'someKey')` and producing a smell in my code. I'm thinking of an interface where all my entities were resolved at once, without defining each of them in my call to `resolve` (they are already defined in my `.l20n` files)... and getting a dictionary back like shown above with all my entities resolved.

zbraniecki:
Sweet! Thanks a lot for bringing this up @jedireza !

tests/node/index.js
stasm: Another way of writing this code could be this:

```javascript
const identifiers = ['whatCanIAsk', 'openApp', 'callNumber', 'notFound'];
Promise.all(identifiers.map(ctx.resolve.bind(ctx, langs)));
```

..which I think is much simpler and also readable.

Regarding the return value of `Promise.all`, I don't think we can really return a dict because of interpolations.  You might want to interpolate different values into the same string, in which case a dict doesn't work.  Consider the following example:

```javascript
const strings = [
  ['aMsgFrom', 'Mary'],
  ['aMsgFrom', 'John']
];
Promise.all(strings.map(([id, args]) => ctx.resolve(langs, id, args)));
```

This also shows how the ES2015 syntax helps us write readable code with `Promise.all`.

tests/node/index.js
stasm: The reason why we currently require `langs` as the first argument are race conditions. It allows us to not have to keep and synchronize `langs` across multiple contexts.  But I'm open to discussion if you have better ideas.

build/webpack.js
```diff
@@ -90,4 +90,16 @@ module.exports = {
       loaders: [babel]
     }
   },
+  nodeBrowser: {
+    context: path('../src'),
+    entry: './runtime/node/browser.js',
```
stasm: I'm not a fan of the naming here.  This isn't `node`, but rather `npm` if I understand correctly?  Also, in other target points we use `web` instead of `browser`.  How about we create `src/runtime/web/common.js` instead of `src/runtime/node/browser.js` and the dist file is called `dist/webcompat/l20n-common.js`?

tests/node/index.js
jedireza: Great point about multiple strings with different parameters.

build/webpack.js
jedireza: Sounds good to me. I can make this change.

stasm:
I didn't know about the `browser` fields in `package.json`, thanks!  Regarding `main`, have you seen ae5c2b2bc0b10216f639f24dd9983f137cec7216? I added `l20n/compat` which would work well with the `prepublish` scrpit.  Would you recommend we rather do this the other way round?  That is, `require('l20n')` wouldn't need `babel/register` and `require('l20n/esnext')` (name TBD) would?

jedireza:
@stasm I've pushed a commit for the renaming we discussed earlier. Good call, the new naming is much better IMO.

> I didn't know about the browser field in package.json, thanks!

Yeah, it's really cool for universal (aka isomorphic) dev patterns.

> Regarding main, have you seen ae5c2b2? I added `l20n/compat` which would work well with the `prepublish` scrpit. Would you recommend we rather do this the other way round? That is, `require('l20n')` wouldn't need `babel/register` and `require('l20n/esnext')` (name TBD) would?

I think we should point `require('l20n')` to the ES5 compatible version (done in this PR). Module consumers shouldn't worry about which style of JavaScript the module is written in since they're just consuming the API. With that said I don't think we'll need to have a `require('l20n/esnext')`. If everyone agrees, I can remove the `./compat.js` file then?

I'd also recommend that we remove mentions of using `require('babel/register')` too. It's not a good practice to encourage something that pollutes the global scope and overwrites __all__ calls to `require`. 
If we all agree, I can also update the docs for node.

I assume we'll remove the `./tests/node/*` files I created. What other changes would you like me to make?

src/runtime/node/io.js
```diff
@@ -2,6 +2,7 @@
 
 import { readFile } from 'fs';
 import { L10nError } from '../../lib/errors';
+import 'string.prototype.endswith';
```
jedireza: FYI: this is required for server side node since it doesn't ship with an `String.prototype.endsWith`. 

It's also worth mentioning, that we may want to include this in our other web distributions since not all browsers are shipping `String.prototype.endsWith` either.

stasm:
> I think we should point require('l20n') to the ES5 compatible version (done in this PR). Module consumers shouldn't worry about which style of JavaScript the module is written in since they're just consuming the API. With that said I don't think we'll need to have a require('l20n/esnext'). If everyone agrees, I can remove the ./compat.js file then?

Just thinking out loud:  if someone already uses `babel/register` in their code or io.js, wouldn't they prefer to `require` an es6 version of the code?

> I'd also recommend that we remove mentions of using require('babel/register') too. It's not a good practice to encourage something that pollutes the global scope and overwrites all calls to require.
If we all agree, I can also update the docs for node.

The node build target doesn't expose all internal APIs on purpose, but I'd still like to make it possible to `require` individual modules, e.g. `require('l20n/src/lib/resolver')`, kind of like we do in https://github.com/l20n/l20n.js/blob/0ff64b20cd9bb7be4aeb343cc08a111d8af7682e/tools/format.js#L9-L10.  Two questions:  do you think this is a good idea and a good practice? And: does it go without saying that you need `babel/register` for this to work?

jedireza:
> Just thinking out loud: if someone already uses `babel/register` in their code or io.js, wouldn't they prefer to `require` an es6 version of the code?

I don't see why. The majority of modules out there are not written in ES6 syntax.

> The node build target doesn't expose all internal APIs on purpose, but I'd still like to make it possible to require individual modules, e.g. `require('l20n/src/lib/resolver')`, kind of like we do in https://github.com/l20n/l20n.js/blob/0ff64b20cd9bb7be4aeb343cc08a111d8af7682e/tools/format.js#L9-L10. Two questions: do you think this is a good idea and a good practice? And: does it go without saying that you need `babel/register` for this to work?

This is a great point. Originally (before this PR) I created a directory locally called `l20n-node` and ran the entire `./src` folder through `babel` and was able to then `require` everything (although code like `String#endsWith` would still throw). However, what seems like the right move is to expose as little as possible to consume the public API.

IMO if/when someone wanted to consumer some of the internals, they'd need to use a live transformer (ex: `babel/register`) or request we expose more in `./src/runtime/node` by default.

tests/node/index.js
stasm: Let's move this discussion to another issue, pr or bug.  This pr already has a lot of good things in it and I'd like to land it soon.

stasm:
> I think we should point require('l20n') to the ES5 compatible version (done in this PR). Module consumers shouldn't worry about which style of JavaScript the module is written in since they're just consuming the API. With that said I don't think we'll need to have a require('l20n/esnext'). If everyone agrees, I can remove the ./compat.js file.

> I'd also recommend that we remove mentions of using require('babel/register') too. It's not a good practice to encourage something that pollutes the global scope and overwrites all calls to require.
If we all agree, I can also update the docs for node.

> I assume we'll remove the ./tests/node/* files I created (which I see are what's making the tests fail on Travis). What other changes would you like me to make?

These are great suggestions, thanks @jedireza.  I especially like how `prepublish` obsoletes the need of `l20n/compat` and I'm glad you chimed in to teach me the proper way :)

tests/node/index.js
jedireza: That works for me.

jedireza:
I've pushed commits and updated the node docs, removed `./compat.js` and removed the demo node script files and resource file fixtures.

package.json
```diff
@@ -17,12 +17,15 @@
   ],
   "bugs": "https://bugzilla.mozilla.org/enter_bug.cgi?product=Firefox%20OS&component=Gaia::L10n",
   "scripts": {
-    "test": "grunt merge-conflict lint test"
+    "test": "grunt merge-conflict lint test",
+    "prepublish": "grunt webpack",
+    "test-node": "node ./tests/node"
```
stasm: Do we still need this?

jedireza: Oh my bad. It's been removed.

stasm:
This looks good to me, r+.  @zbraniecki, do you want to take a look as well?

jedireza:
I'm testing the web+commonjs distribution locally right now. Let me also make sure that's working before we merge.

zbraniecki:
r+ - I'm not really deep in the web stack right now. But it seems to look good, I trust Reza and it doesn't seem to affect gaia build.

stasm:
OK, @jedireza, let me know when you think this is ready to merge, and thanks so much for working on this!

jedireza:
Besides what having to pass the `src` attribute with the lang objects for browserify/webpack apps, this is working as I expected.

stasm:
Do you want to update the docs or should we fix this in another pr?

jedireza:
Since the issue isn't with the node implementation, but specifically for the web+commonjs distribution, I'm unsure where I should add a note about this. Maybe best to handle in another PR?

stasm:
This is because webruntime's io uses `src` https://github.com/l20n/l20n.js/blob/90bf5b66fcae4150c48ab875a4e380d5709c71ff/src/runtime/web/io.js#L64 and node's doesn't: https://github.com/l20n/l20n.js/blob/90bf5b66fcae4150c48ab875a4e380d5709c71ff/src/runtime/node/io.js#L19.

OK, I see, this doesn't really belong in `docs/node`.  You're right, let's fix this in a separate issue.

PR #67, opened 2015-08-04
Remove support for .innerHTML

stasm:
@zbraniecki r?

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -1,7 +1,5 @@
 'use strict';
 
-import { L10nError } from '../../lib/errors';
-
 // match the opening angle bracket (<) in HTML tags, and HTML entities like
 // &amp;, &#0038;, &#x0026;.
 const reOverlay = /<|&#?\w+;/;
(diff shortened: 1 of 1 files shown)


zbraniecki:
r+ unless it affects l10n.js apps at build time.

stasm:
Great point, I double checked and there is indeed one string that might be affected: https://github.com/mozilla-b2g/gaia/blob/f7762c4b56b881b1569671973d02754032ed4685/apps/ftu/locales/ftu.en-US.properties#L203

It should be enough to just remove `.innerHTML` from it in the Gaia pr. I'm testing right now, though, and it seems to work either way.

stasm:
OK,so what I think I happens is:  build/l10n/l20n.js clears the fxa-upsell node during the pretranslation, but FTU still has its own logic which uses `setAttributes`and localizes the node with the runtime l10n.js, which supports innerHTML.

stasm:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1191163 to remove that `innerHTML` from FTU before I merge this.

PR #68, opened 2015-08-05
using docker builds on travis

jedireza:
If we're not using `sudo` for anything we can turn it off and get to use the faster docker containers offered by Travis.

http://docs.travis-ci.com/user/workers/container-based-infrastructure/


jedireza:
:open_mouth: Wow. That finished fast!

stasm:
Nice.

PR #69, opened 2015-08-05
setAttributes can take an object of Promises

stasm:
Instead of doing this:

```javascript
Promise.all([oneAsyncThing(), otherAsyncThing()]).then(
  (oneValue, otherValue) => document.l10n.setAttributes(
    elem, 'foo', { one: oneValue, other: otherValue }));
```

…this PR makes it possible to do this:

```javascript
document.l10n.setAttributes(
    elem, 'foo', { one: oneAsyncThing(), other: otherAsyncThing() }));
```

@zbraniecki, what do you think?

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -29,11 +29,25 @@ const allowed = {
   }
 };
 
+function zip(keys, values) {
+  return keys.reduce(
+    (seq, cur, i) => { seq[cur] = values[i]; return seq; },
+    Object.create(null));
+}
+
 export function setAttributes(element, id, args) {
-  element.setAttribute('data-l10n-id', id);
-  if (args) {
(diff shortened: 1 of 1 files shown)


stasm:
One limitation is that Promises can only be the top-level members of the `args` object.

So this is okay:

```javascript
document.l10n.setAttributes(
    elem, 'hello', { user: Promise.resolve({ name: 'Jamie' }) }));
```

But this isn't:

```javascript
document.l10n.setAttributes(
    elem, 'hello', { user: { name: Promise.resolve('Jamie') }}));
```

zbraniecki:
I think I'd prefer to keep setAttributes simple. The main reason we have a promise passed to args is because we need to formatValue that is then passed to setAttributes.

We should replace it in l20n with entity references from inside of l20n entity.

stasm:
Sounds good to me!

PR #71, opened 2015-08-06
Bug 1190539 - Escape HTML in args

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1190539

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -3,6 +3,12 @@
 // match the opening angle bracket (<) in HTML tags, and HTML entities like
 // &amp;, &#0038;, &#x0026;.
 const reOverlay = /<|&#?\w+;/;
+const reHtml = /[&<>]/g;
+const htmlEntities = {
+  '&': '&amp;',
+  '<': '&lt;',
+  '>': '&gt;',
+};
 
 const allowed = {
   elements: [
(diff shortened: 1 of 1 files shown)


src/bindings/html/dom.js
```diff
@@ -106,10 +116,12 @@ function camelCaseToDashed(string) {
 }
 
 function getElementTranslation(view, langs, elem) {
-  const l10n = getAttributes(elem);
+  const id = elem.getAttribute('data-l10n-id');
+  const args = JSON.parse(
+    escapeHtml(elem.getAttribute('data-l10n-args')));
```
zbraniecki: does it make sense here to `if (args)` so that we don't run JSON.parse when we don't have args?

stasm: Yeah, I updated the whole thing to handle falsy `id` and `args` better.

src/bindings/html/dom.js
```diff
@@ -106,10 +112,21 @@ function camelCaseToDashed(string) {
...
+  const id = elem.getAttribute('data-l10n-id');
+
+  if (!id) {
+    return false;
+  }
+
+  const args = elem.getAttribute('data-l10n-args');
+
+  if (!args) {
+    return view.ctx.resolve(langs, id);
```
stasm: This early return has the benefit of guarding against a `TypeError` if `data-l10n-args` is an empty string.

PR #72, opened 2015-08-17
Bug 1195539 - Filter out non-l10n elements from translateMutations and remove condition race

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1195539

### DIFF
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -48,15 +48,14 @@ export function getAttributes(element) {
 }
 
 function getTranslatables(element) {
-  const nodes = [];
+  const nodes = Array.from(element.querySelectorAll('[data-l10n-id]'));
 
   if (typeof element.hasAttribute === 'function' &&
       element.hasAttribute('data-l10n-id')) {
     nodes.push(element);
   }
 
-  return nodes.concat.apply(
-    nodes, element.querySelectorAll('*[data-l10n-id]'));
+  return nodes;
 }
 
 export function translateMutations(view, langs, mutations) {
@@ -70,7 +69,11 @@ export function translateMutations(view, langs, mutations) {
       case 'childList':
         for (let addedNode of mutation.addedNodes) {
           if (addedNode.nodeType === addedNode.ELEMENT_NODE) {
-            targets.add(addedNode);
+            if (addedNode.childElementCount) {
+              getTranslatables(addedNode).forEach(targets.add.bind(targets));
+            } else {
+              targets.add(addedNode);
+            }
           }
         }
         break;
@@ -81,21 +84,11 @@ export function translateMutations(view, langs, mutations) {
     return;
   }
 
-  const elements = [];
-
-  targets.forEach(target => target.childElementCount ?
-      elements.push(...getTranslatables(target)) : elements.push(target));
-
-  Promise.all(
-    elements.map(elem => getElementTranslation(view, langs, elem))).then(
-      translations => applyTranslations(view, elements, translations));
+  translateElements(view, langs, Array.from(targets));
 }
 
 export function translateFragment(view, langs, frag) {
-  const elements = getTranslatables(frag);
-  return Promise.all(
-    elements.map(elem => getElementTranslation(view, langs, elem))).then(
-      translations => applyTranslations(view, elements, translations));
+  return translateElements(view, langs, getTranslatables(frag));
 }
 
 function camelCaseToDashed(string) {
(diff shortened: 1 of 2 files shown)


src/bindings/html/dom.js
```diff
@@ -42,15 +42,14 @@ export function getAttributes(element) {
 }
 
 function getTranslatables(element) {
-  const nodes = [];
+  const nodes = [...element.querySelectorAll('*[data-l10n-id]')];
```
stasm: I really like this way of transforming a `NodeList` into an array! It does seem a bit slower than the traditional methods (cf. http://jsperf.com/qsa-slice-vs-spread) but I'm all for keeping it here.

src/bindings/html/dom.js
```diff
@@ -42,15 +42,14 @@ export function getAttributes(element) {
...
 
-  if (typeof element.hasAttribute === 'function' &&
+  if (element.nodeType !== element.DOCUMENT_FRAGMENT_NODE &&
```
stasm: Why this change?  The original served the same purpose and was more generic and possibly future-proof.

src/bindings/html/dom.js
```diff
@@ -77,19 +76,31 @@ export function translateMutations(view, langs, mutations) {
 
   const elements = [];
 
-  targets.forEach(target => target.childElementCount ?
-      elements.push(...getTranslatables(target)) : elements.push(target));
+  targets.forEach(target => {
+    if (target.childElementCount) {
+      elements.push(...getTranslatables(target));
+    } else {
+      if (target.hasAttribute('data-l10n-id')) {
```
stasm: You can collapse this into a `else if`.  I think I preferred the ternary version but that's just style.

zbraniecki: the ternary style would require me to have an if in the else structure. I feel like we're paying with code readability for dense coding at some point. if/else block may be longer but I feel like it's more readable. Would you agree?

src/bindings/html/dom.js
zbraniecki: I think I thought that it's smarter... but I'm not sure anymore. Will revert.

src/bindings/html/dom.js
stasm: It all boils down to personal preference, and I'm okay with your approach.

I think the following is also quite readable and I like how it's a single expression.

```javascript
targets.forEach(target => target.childElementCount ?
  elements.push(...getTranslatables(target)) :
  target.hasAttribute('data-l10n-id') ?
    elements.push(target)) : undefined
```

Again, this is just personal preference.  r+ on your version! :)

src/bindings/html/dom.js
```diff
@@ -77,19 +76,31 @@ export function translateMutations(view, langs, mutations) {
...
+  if (elements.length) {
+    translateElements(view, langs, elements);
+  }
+}
 
-  Promise.all(
-    elements.map(elem => getElementTranslation(view, langs, elem))).then(
-      translations => applyTranslations(view, elements, translations));
+function translateElements(view, langs, elements) {
+  return view.ready.then(() => {
```
stasm: I'm still thinking about this :)

src/bindings/html/dom.js
stasm: Also, and it's just a nit:  I think we can remove the `*` from the qSA.

src/bindings/html/dom.js
stasm: I like that we're not repeating ourselves here and that you abstracted `translateElements` into its own function.  Can you move it closer to where `translateElement` is defined?

However, I think that the `view.ctx.fetch` guard should be set in view.js, not here.  In dom.js, we make certain assumptions and one of them is that the view takes care of guaranteeing that the first language has been loaded.

(We can discuss if this assumption is valid.  For now, I suggest we keep it and keep things consistent.)

Take a look at `translateFragment` in dom.js and in view.js.  The view calls `ctx.fetch` before it calls dom's `translateFragment`.  I think we should do the same for mutations:  in view's `onMutations` we should add `ctx.fetch`, just like it's done a few lines above in `translateFragment`.

What do you think?

zbraniecki: Great! Updated the patch.

src/bindings/html/dom.js
```diff
@@ -42,15 +42,14 @@ export function getAttributes(element) {
 }
 
 function getTranslatables(element) {
-  const nodes = [];
+  const nodes = [...element.querySelectorAll('[data-l10n-id]')];
 
-  if (typeof element.hasAttribute === 'function' &&
+  if (element.nodeType !== element.DOCUMENT_FRAGMENT_NODE &&
```
stasm: Forgot to revert this?

src/bindings/html/dom.js
stasm: Can you still collapse the `else { if` into an `else if`?

src/bindings/html/dom.js
```diff
@@ -77,19 +76,30 @@ export function translateMutations(view, langs, mutations) {
...
+    }
+  });
 
-  Promise.all(
+  if (elements.length) {
+    translateElements(view, langs, elements);
+  }
+}
+
+function translateElements(view, langs, elements) {
```
stasm: And put this closer to `translateElement`, please?

PR #73, opened 2015-08-18
Bug 1196009 - Bring back View.formatValue

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1196009

### DIFF
--- a/src/lib/context.js
+++ b/src/lib/context.js
@@ -22,7 +22,7 @@ export class Context {
   }
 
   _formatEntity(lang, args, entity, id) {
-    const [, value] = this._formatTuple.call(this, lang, args, entity, id);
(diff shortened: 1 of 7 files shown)


src/lib/context.js
```diff
@@ -53,7 +57,7 @@ export class Context {
           () => langs);
   }
 
-  resolve(langs, id, args) {
+  _resolve(langs, id, args, formatter) {
     const lang = langs[0];
 
     if (!lang) {
```
stasm: I can't put the comment under the specific line, but I think we'll need to differentiate here between value and object when we return `return { value: id, attrs: null };`.

zbraniecki: I fixed that

PR #74, opened 2015-08-19
Split webpack entry points over src/runtime/**/config.js files

stasm:
This is a simple change that automates the definitions of webpack tasks a little bit. Between bindings, runtimes and webpack entry points, figuring out what to edit might get a bit confusing.  In this PR I'm introducing a mechanism which looks for `config.js` files in `src/runtime` and creates grunt tasks based on the config data in them.


stasm:
@zbraniecki r?

zbraniecki:
r+

PR #77, opened 2015-08-20
Make document.l10n available synchronously

zbraniecki:


src/runtime/web/index.js
```diff
@@ -9,6 +9,7 @@ const readyStates = {
   interactive: 1,
   complete: 2
 };
+const service = new Service(fetch);
```
stasm: Can you create this in `init`?

zbraniecki: yup!

stasm:
r+, thanks!

PR #78, opened 2015-09-11
Bug 1204086 - Add formatValues API

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1204086

### DIFF
--- a/src/bindings/gaiabuild/view.js
+++ b/src/bindings/gaiabuild/view.js
@@ -57,8 +57,8 @@ export class View {
   observe() {}
   disconnect() {}
 
-  _resolveEntity(langs, id, args) {
-    return this.ctx.resolveEntity(langs, id, args);
+  _resolveEntities(langs, keys) {
+    return this.ctx.resolveEntities(langs, keys);
   }
 
   translateDocument(code) {
--- a/src/bindings/html/dom.js
+++ b/src/bindings/html/dom.js
@@ -48,7 +48,9 @@ export function translateMutations(view, langs, mutations) {
             if (addedNode.childElementCount) {
               getTranslatables(addedNode).forEach(targets.add.bind(targets));
             } else {
-              targets.add(addedNode);
+              if (addedNode.hasAttribute('data-l10n-id')) {
+                targets.add(addedNode);
+              }
             }
           }
         }
@@ -67,51 +69,27 @@ export function translateFragment(view, langs, frag) {
   return translateElements(view, langs, getTranslatables(frag));
 }
 
-function getElementTranslation(view, langs, elem) {
-  const id = elem.getAttribute('data-l10n-id');
+function getElementsTranslation(view, langs, elems) {
+  const keys = elems.map(elem => {
+    const id = elem.getAttribute('data-l10n-id');
+    const args = elem.getAttribute('data-l10n-args');
+    return args ? [
+      id,
+      JSON.parse(args.replace(reHtml, match => htmlEntities[match]))
+    ] : id;
+  });
 
-  if (!id) {
-    return false;
-  }
-
-  const args = elem.getAttribute('data-l10n-args');
-
-  if (!args) {
-    return view._resolveEntity(langs, id);
-  }
-
-  return view._resolveEntity(
-    langs, id, JSON.parse(
-      args.replace(reHtml, match => htmlEntities[match])));
-}
-
-export function translateElement(view, langs, elem) {
-  return getElementTranslation(view, langs, elem).then(
-    translation => applyTranslation(view, elem, translation));
+  return view._resolveEntities(langs, keys);
 }
 
 function translateElements(view, langs, elements) {
-  return Promise.all(
-    elements.map(elem => getElementTranslation(view, langs, elem))).then(
-      translations => applyTranslations(view, elements, translations));
-}
-
-function applyTranslation(view, elem, translation) {
-  if (!translation) {
-    return false;
-  }
-
-  view.disconnect();
-  overlayElement(elem, translation);
-  view.observe();
+  return getElementsTranslation(view, langs, elements).then(
+    translations => applyTranslations(view, elements, translations));
 }
 
 function applyTranslations(view, elems, translations) {
   view.disconnect();
   for (let i = 0; i < elems.length; i++) {
-    if (translations[i] === false) {
-      continue;
-    }
     overlayElement(elems[i], translations[i]);
   }
   view.observe();
--- a/src/bindings/html/view.js
+++ b/src/bindings/html/view.js
@@ -48,13 +48,19 @@ export class View {
     return this.service.env.emit(...args);
   }
 
-  _resolveEntity(langs, id, args) {
-    return this.service.resolveEntity(this, langs, id, args);
+  _resolveEntities(langs, keys) {
+    return this.service.resolveEntities(this, langs, keys);
   }
 
   formatValue(id, args) {
     return this.service.initView(this).then(
-      langs => this.service.resolveValue(this, langs, id, args));
+      langs => this.service.resolveValues(this, langs, [[id, args]])).then(
+        values => values[0]);
+  }
+
+  formatValues(...keys) {
+    return this.service.initView(this).then(
+      langs => this.service.resolveValues(this, langs, keys));
   }
 
   translateFragment(frag) {
--- a/src/lib/context.js
+++ b/src/lib/context.js
@@ -57,39 +57,47 @@ export class Context {
           () => langs);
   }
 
-  _resolve(langs, id, args, formatter) {
+  _resolve(langs, keys, formatter, prevResolved) {
     const lang = langs[0];
 
     if (!lang) {
-      this._env.emit('notfounderror', new L10nError(
-        '"' + id + '"' + ' not found in any language', id), this);
-      if (formatter === this._formatEntity) {
-        return { value: id, attrs: null };
-      } else {
-        return id;
-      }
+      return reportMissing.call(this, keys, formatter, prevResolved);
     }
 
-    const entity = this._getEntity(lang, id);
+    let hasUnresolved = false;
 
-    if (entity) {
-      return Promise.resolve(
-        formatter.call(this, lang, args, entity, id));
-    } else {
-      this._env.emit('notfounderror', new L10nError(
-        '"' + id + '"' + ' not found in ' + lang.code, id, lang), this);
+    const resolved = keys.map((key, i) => {
+      if (prevResolved && prevResolved[i] !== undefined) {
+        return prevResolved[i];
+      }
+      const [id, args] = Array.isArray(key) ?
+        key : [key, undefined];
+      const entity = this._getEntity(lang, id);
+
+      if (entity) {
+        return formatter.call(this, lang, args, entity, id);
+      }
+
+      this._env.emit('notfounderror',
+        new L10nError('"' + id + '"' + ' not found in ' + lang.code,
+          id, lang), this);
+      hasUnresolved = true;
+    });
+
+    if (!hasUnresolved) {
+      return Promise.resolve(resolved);
     }
 
     return this.fetch(langs.slice(1)).then(
-      nextLangs => this._resolve(nextLangs, id, args, formatter));
+      nextLangs => this._resolve(nextLangs, keys, formatter, resolved));
   }
 
-  resolveEntity(langs, id, args) {
-    return this._resolve(langs, id, args, this._formatEntity);
+  resolveEntities(langs, keys) {
+    return this._resolve(langs, keys, this._formatEntity);
   }
 
-  resolveValue(langs, id, args) {
-    return this._resolve(langs, id, args, this._formatValue);
+  resolveValues(langs, keys) {
+    return this._resolve(langs, keys, this._formatValue);
   }
 
   _getEntity(lang, id) {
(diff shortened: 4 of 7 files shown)


src/lib/context.js
```diff
@@ -84,6 +84,35 @@ export class Context {
...
+        ret[id] = this._formatValue(lang, idArgsKVP[id], entity, id);
+      } else {
+        this._env.emit('notfounderror',
+          new L10nError('"' + id + '"' + ' not found in ' + lang.code,
+            id, lang), this);
+      }
+    }
+
+    return this.fetch(langs.slice(1)).then(
+      nextLangs => this._resolveManyValues(nextLangs, idArgsKVP));
```
stasm: Where is `_resolveManyValues` defined?

src/lib/context.js
```diff
@@ -84,6 +84,35 @@ export class Context {
...
+    }
+
+    var ret = Object.create(null);
+
+    for (var i = 0; i < ids.length; i++) {
+      const id = ids[i];
+      const entity = this._getEntity(lang, id);
+
+      if (entity) {
+        ret[id] = this._formatValue(lang, idArgsKVP[id], entity, id);
```
stasm: Where do you return `ret`?

src/lib/context.js
zbraniecki: it should be without '_'. I'll fix it with the next patch

src/bindings/html/dom.js
```diff
@@ -67,51 +67,39 @@ export function translateFragment(view, langs, frag) {
...
+  return getElementsTranslation(view, langs, elements).then(
+    translations => applyTranslations(view, elements, translations));
 }
 
 function applyTranslations(view, elems, translations) {
   view.disconnect();
   for (let i = 0; i < elems.length; i++) {
-    if (translations[i] === false) {
-      continue;
-    }
```
zbraniecki: there's no possibility for it to be false

src/bindings/html/dom.js
```diff
@@ -67,51 +67,39 @@ export function translateFragment(view, langs, frag) {
...
+      idArgsList.push(id);
+    } else {
+      idArgsList.push([
+        id,
+        JSON.parse(args.replace(reHtml, match => htmlEntities[match]))
+      ]);
+    }
+  }
 
-export function translateElement(view, langs, elem) {
```
zbraniecki: this is not called anywhere

src/lib/context.js
```diff
@@ -57,39 +57,67 @@ export class Context {
...
+      } else {
+        this._env.emit('notfounderror',
+          new L10nError('"' + id + '"' + ' not found in ' + lang.code,
+            id, lang), this);
+        hasUnresolved = true;
+      }
+    });
+
+    if (!hasUnresolved) {
+      return Promise.all(resolved);
```
stasm: Is `resolved` an array of promises?  `formatter` should return synchronous values, in which case you should be able to `return Promise.resolve(resolved)` here.  Note that this should much faster:  it's only one promise vs. one promise per element of `resolved`.

src/bindings/html/view.js
```diff
@@ -48,13 +48,19 @@ export class View {
...
   }
 
   formatValue(id, args) {
     return this.service.initView(this).then(
-      langs => this.service.resolveValue(this, langs, id, args));
+      langs => this.service.resolveValues(this, langs, [[id, args]])).then(
+        values => values[0]);
+  }
+
+  formatValues(idArgsList) {
```
stasm: `...idArgsList`?

src/bindings/html/dom.js
```diff
@@ -67,51 +69,39 @@ export function translateFragment(view, langs, frag) {
...
 
-function getElementTranslation(view, langs, elem) {
-  const id = elem.getAttribute('data-l10n-id');
+function getElementsTranslation(view, langs, elems) {
+  var idArgsList = [];
 
-  if (!id) {
-    return false;
-  }
+  for (let elem of elems) {
```
stasm: IIRC we wanted to avoid `for..of` for now because of how Babel transpiles it using generators, which in turn require a polyfill in older browsers.

src/bindings/html/dom.js
```diff
@@ -67,51 +69,39 @@ export function translateFragment(view, langs, frag) {
...
+    const args = elem.getAttribute('data-l10n-args');
 
-  return view._resolveEntity(
-    langs, id, JSON.parse(
-      args.replace(reHtml, match => htmlEntities[match])));
-}
+    if (!args) {
+      idArgsList.push(id);
+    } else {
+      idArgsList.push([
```
stasm: Maybe move the `!args` check into a ternary `if` inside of `push`?

src/lib/context.js
```diff
@@ -57,39 +57,67 @@ export class Context {
...
-      if (formatter === this._formatEntity) {
-        return { value: id, attrs: null };
-      } else {
-        return id;
-      }
+      return this._reportMissing(idArgsList, formatter, resolved);
     }
 
-    const entity = this._getEntity(lang, id);
+    var hasUnresolved = false;
```
stasm: `let`?

src/lib/context.js
```diff
@@ -57,39 +57,67 @@ export class Context {
...
+
+    idArgsList.forEach((idArgs, i) => {
+      if (i in resolved) {
+        return;
+      }
+      const id = Array.isArray(idArgs) ? idArgs[0] : idArgs;
+      const args = Array.isArray(idArgs) ? idArgs[1] : undefined;
```
stasm: Maybe use destructuring to avoid the double `isArray` check?

const [id, args] = Array.isArray(idArgs) ? idArgs : [idArgs, undefined];

stasm: Also, maybe call it `key` instead of `idArgs`? It can be a simple identifier, or a id+args tuple.  The `'key` name would work particularly well if we used a Map as the return value.

zbraniecki: love it!

src/lib/context.js
zbraniecki: I went looking for a facepalm emoji only to learn that it's considered for inclusion in Unicode 9.0. 
I'm very disappointed.

src/bindings/html/dom.js
```diff
@@ -48,7 +48,9 @@ export function translateMutations(view, langs, mutations) {
             if (addedNode.childElementCount) {
               getTranslatables(addedNode).forEach(targets.add.bind(targets));
             } else {
-              targets.add(addedNode);
+              if (addedNode.hasAttribute('data-l10n-id')) {
```
stasm: Wouldn't this break if `addedNode` is a childless `DocumentFragment`?

src/bindings/html/dom.js
```diff
@@ -67,51 +69,30 @@ export function translateFragment(view, langs, frag) {
...
-  }
+  elems.forEach(elem => {
+    const id = elem.getAttribute('data-l10n-id');
+    const args = elem.getAttribute('data-l10n-args');
 
-  return view._resolveEntity(
-    langs, id, JSON.parse(
-      args.replace(reHtml, match => htmlEntities[match])));
-}
+    keys.push( args ? [
```
stasm: This looks like you'd be better of using `elems.map`.

src/bindings/html/dom.js
```diff
@@ -67,51 +69,30 @@ export function translateFragment(view, langs, frag) {
...
+  return view._resolveEntities(langs, keys);
 }
 
 function translateElements(view, langs, elements) {
-  return Promise.all(
-    elements.map(elem => getElementTranslation(view, langs, elem))).then(
-      translations => applyTranslations(view, elements, translations));
-}
-
-function applyTranslation(view, elem, translation) {
```
stasm: I *really* like that we're making the code simpler by removing `translateElement` and `applyTranslation`!

src/bindings/html/view.js
```diff
@@ -48,13 +48,19 @@ export class View {
     return this.service.env.emit(...args);
   }
 
-  _resolveEntity(langs, id, args) {
-    return this.service.resolveEntity(this, langs, id, args);
+  _resolveEntities(langs, idArgsList) {
+    return this.service.resolveEntities(this, langs, idArgsList);
```
stasm: Let's use `keys` here to be consistent.

src/lib/context.js
```diff
@@ -57,39 +57,67 @@ export class Context {
           () => langs);
   }
 
-  _resolve(langs, id, args, formatter) {
+  _reportMissing(missingKeys, formatter, resolved) {
```
stasm: Maybe define this as a module function instead of a method?

stasm: Also, are those really the `missingKeys` or just `keys`?

src/bindings/html/dom.js
zbraniecki: Inserting an empty DocumentFragment does not trigger mutation observer, and we guard against non-element nodes.

src/lib/context.js
```diff
@@ -57,39 +57,47 @@ export class Context {
...
-        return id;
-      }
+      return reportMissing.call(this, keys, formatter, resolved);
     }
 
-    const entity = this._getEntity(lang, id);
+    let hasUnresolved = false;
+
+    keys.forEach((key, i) => {
+      if (i in resolved) {
```
stasm: What do you think about using `map` here as well?

    const resolved = keys.map((key, i) => {
      if (prevResolved !== undefined) {
        // …
    });

tests/lib/context/basic_test.js
```diff
@@ -15,6 +15,12 @@ describe('A simple context with one resource', function() {
   describe('resolveEntity', function() {
     var env, ctx;
 
+    var resolveEntity = function(langs, id, args) {
+      var entityPromise = args ? ctx.resolveEntities(langs, [[id, args]]) :
+        ctx.resolveEntities(langs, [id]);
+      return entityPromise.then(([value]) => value);
```
stasm: Not sure if I like this.  Can't we just use `resolveEntities` in the tests?

src/lib/context.js
zbraniecki: it would make resolved be recreated, instead of filled, which would increase mem/GC, am I wrong?

stasm: Well, GC helps :).  I was wondering about sparse arrays.  Wouldn't it be cleaner to have a constant length array here at all times?

zbraniecki: GC costs CPU. I would prefer to avoid changing algorithm that is better optimized for one that requires more from the engine. I don't think it would be cleaner since the resolved array matches the keys array and they zip well together and I'm just iterating over languages (if I need to), but keeping those two by reference.

src/bindings/html/dom.js
```diff
@@ -67,51 +69,27 @@ export function translateFragment(view, langs, frag) {
...
-  return view._resolveEntity(
-    langs, id, JSON.parse(
-      args.replace(reHtml, match => htmlEntities[match])));
-}
-
-export function translateElement(view, langs, elem) {
-  return getElementTranslation(view, langs, elem).then(
-    translation => applyTranslation(view, elem, translation));
+function getElementsTranslation(view, langs, elems) {
+  var keys = elems.map(elem => {
```
stasm: `const`?

src/lib/context.js
```diff
@@ -57,39 +57,48 @@ export class Context {
...
-        '"' + id + '"' + ' not found in ' + lang.code, id, lang), this);
+      if (entity) {
+        return formatter.call(this, lang, args, entity, id);
+      }
+
+      this._env.emit('notfounderror',
+        new L10nError('"' + id + '"' + ' not found in ' + lang.code,
+          id, lang), this);
+      hasUnresolved = true;
+      return undefined;
```
stasm: I don't think that's needed here, since the function returns `undefined` implicitly.  Up to you.

src/bindings/gaiabuild/view.js
```diff
@@ -57,8 +57,8 @@ export class View {
   observe() {}
   disconnect() {}
 
-  _resolveEntity(langs, id, args) {
-    return this.ctx.resolveEntity(langs, id, args);
+  _resolveEntities(langs, keys) {
+    return this.ctx.resolveEntity(langs, keys);
```
stasm: Typo: `resolveEntities` here.

src/lib/context.js
```diff
@@ -57,39 +57,48 @@ export class Context {
...
+    });
+
+    if (!hasUnresolved) {
+      return Promise.resolve(resolved);
     }
 
-    return this.fetch(langs.slice(1)).then(
-      nextLangs => this._resolve(nextLangs, id, args, formatter));
+    return this.fetch(langs.slice(1)).then(nextLangs =>
+      this._resolve(nextLangs, keys, formatter, resolved));
```
stasm: Nit: for one-liners, I try to keep the argument and the arrow in the same line as the expression.  (Like it used to be here.)

tests/lib/context/basic_test.js
```diff
@@ -113,15 +158,49 @@ describe('A simple context with one resource', function() {
...
+          assert.strictEqual(about, i('About Firefox', 'Firefox'));
+        }
+      ).then(done, done);
+    });
+
+    it('should return for many including an id/args pair', function(done) {
+      var args = {unread: 1};
+      ctx.resolveValues(langs, [
+        'brandName',
+        ['unreadMessages', args],
```
stasm: Could you add one more key here with the same id?

    ['unreadMessages', {unread: 1}],
    ['unreadMessages', {unread: 2}],

PR #79, opened 2015-09-14
Bug 1204067 - Expose qps

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1204067

### DIFF
--- a/src/bindings/html/view.js
+++ b/src/bindings/html/view.js
@@ -1,6 +1,7 @@
 'use strict';
 
-import { getResourceLinks } from '../../bindings/html/head';
+import { qps } from '../../lib/pseudo';
+import { getResourceLinks } from './head';
 import {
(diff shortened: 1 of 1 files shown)


src/bindings/html/view.js
```diff
@@ -16,6 +17,7 @@ const observerConfig = {
 export class View {
   constructor(doc) {
     this.doc = doc;
```
zbraniecki: just a thought as I saw it. Why is View.doc public? Does document.l10n.doc make any sense? Should it be at lest prefixed with '_'?

stasm: Yeah, I haven't figured out how to deal with private members in this code yet.  Should we also have `_service`, `_observe` and `_disconnect`?  Or maybe put most of the method definitions in the constructor and use the closure to hide private members?

PR #80, opened 2015-09-14
Bug 1204660 - Introduce a temporary proxy for l20n.js to mimic l10n.js API

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1204660

### DIFF
--- a/src/runtime/gaia/index.js
+++ b/src/runtime/gaia/index.js
@@ -1,0 +1,17 @@
+'use strict';
+
+import '../web/index';
+
+//Bug 1204660 - Temporary proxy for shared code. Will be removed once
+//              l10n.js migration is completed.
+navigator.mozL10n = {
+  setAttributes: document.l10n.setAttributes,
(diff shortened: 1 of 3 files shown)


src/runtime/gaia/index.js
```diff
@@ -0,0 +1,49 @@
...
+
+document.l10n = new View(document);
+
+//Bug 1204660 - Temporary proxy for shared code. Will be removed once
+//              l10n.js migration is completed.
+navigator.mozL10n = {
+  setAttributes: document.l10n.setAttributes,
+  getAttributes: document.l10n.getAttributes,
+  formatValue: document.l10n.formatValue.bind(document.l10n),
+  translateFragment: document.l10n.translateFragment.bind(document.l10n),
```
stasm: How about using arrow function here which I still are still faster than `bind`?

    formatValue: (...args) => document.l10n.formatValue(...args),

PR #83, opened 2015-09-22
Bug 1207120 - Make all public View methods available during DOM loading

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1207120

### DIFF
--- a/src/bindings/html/view.js
+++ b/src/bindings/html/view.js
@@ -1,7 +1,7 @@
 'use strict';
 
 import { qps } from '../../lib/pseudo';
-import { getResourceLinks } from './head';
+import { getResourceLinks, documentReady } from './head';
 import {
   setAttributes, getAttributes, translateFragment, translateMutations
 } from './dom';
@@ -15,9 +15,12 @@ const observerConfig = {
 };
 
 export class View {
-  constructor(doc) {
+  constructor(client, doc) {
     this.doc = doc;
     this.qps = qps;
+
+    this.interactive = documentReady().then(
+      () => init(this, client));
 
     this.ready = new Promise(function(resolve) {
       const viewReady = function(evt) {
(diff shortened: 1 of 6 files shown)


src/runtime/web/index.js
```diff
@@ -4,34 +4,9 @@ import { fetch } from './io';
...
-
-  document.l10n.init(service);
-  document.l10n.languages = navigator.languages;
-}
-
-document.l10n = new View(document);
-
-whenInteractive(init);
+document.l10n = new View(service, document);
+document.l10n.requestLanguages(navigator.languages);
```
zbraniecki: woohoo. That looks beautiful

src/runtime/web/service.js
```diff
@@ -1,39 +1,45 @@
 'use strict';
```
zbraniecki: why are you moving service to runtime? I thought that runtime is supposed to be code that executes, and Service is still a class with no running code?

src/bindings/html/view.js
```diff
@@ -74,20 +73,22 @@ export class View {
...
 }
 
-function translateDocument(langs) {
-  const [view, doc] = [this, this.doc];
+export function translateDocument(view, langs) {
+  const doc = view.doc;
 
   if (langs[0].code === doc.documentElement.getAttribute('lang')) {
-    return Promise.resolve.then(
+    return Promise.resolve().then(
```
zbraniecki: what's this change about? Was that a spelling before?

stasm: It was a typo from 6bf00f35c18b870346eae4708975635335c10b4e, my bad.

src/runtime/web/service.js
stasm: `runtime/web/index` still plugs the service in and starts the listeners. The instance itself is inert except for the language negotiation part.  I moved it to `runtime/` because I think that new runtimes will need their own services.  For instance, I imagine `runtime/nga` which could have a bridge.js-enabled service.js and a corresponding client.js.  You'd then pass the client instance to the view. Some runtimes won't need them at all (e.g. gaiabuild).

There's still some cleaning up to do:  service.js now requires stuff from bindings/html like langnego functions.  I think we can clean this up incrementally as we go.

PR #85, opened 2015-09-28
Bug 1208800 - Format integers passed as arguments

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1208800

### DIFF
--- a/src/lib/context.js
+++ b/src/lib/context.js
@@ -8,6 +8,7 @@ export class Context {
   constructor(env, resIds) {
     this._env = env;
     this._resIds = resIds;
+    this._numberFormatters = null;
   }
 
   _formatTuple(lang, args, entity, id, key) {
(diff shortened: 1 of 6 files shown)


src/lib/context.js
```diff
@@ -8,6 +8,21 @@ export class Context {
   constructor(env, resIds) {
     this._env = env;
     this._resIds = resIds;
+    this._numberFormatters = null;
+  }
+
+  _getNumberFormatter(lang) {
```
stasm: nit: can you move this to the bottom of the file, closer to `_getEntity`, please?

src/lib/resolver.js
```diff
@@ -70,12 +70,14 @@ function subPlaceable(locals, ctx, lang, args, id) {
   try {
     res = resolveIdentifier(ctx, lang, args, id);
```
stasm: Perhaps you could `const [newLocals, value] = resolveIndentifier(…);` here to avoid `res[1]` below?

PR #86, opened 2015-09-29
DOMRetranslated

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1209363

### DIFF
--- a/src/bindings/html/view.js
+++ b/src/bindings/html/view.js
@@ -14,6 +14,8 @@ const observerConfig = {
   attributeFilter: ['data-l10n-id', 'data-l10n-args']
 };
 
+const readiness = new WeakMap();
+
 export class View {
   constructor(client, doc) {
     this._doc = doc;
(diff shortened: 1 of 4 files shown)


src/bindings/html/view.js
```diff
@@ -85,28 +79,42 @@ function onMutations(mutations) {
 }
 
 export function translateDocument(view, langs) {
-  const doc = view._doc;
+  const html = view._doc.documentElement;
 
-  if (langs[0].code === doc.documentElement.getAttribute('lang')) {
-    return Promise.resolve().then(
-      () => dispatchEvent(doc, 'DOMLocalized', langs));
+  if (readiness.get(html)) {
```
zbraniecki: `.has`?

src/bindings/html/view.js
```diff
@@ -14,6 +14,8 @@ const observerConfig = {
   attributeFilter: ['data-l10n-id', 'data-l10n-args']
 };
 
+const readiness = new WeakMap();
```
zbraniecki: can it be a `WeakSet`?

stasm: Ha, it could, if only it was better supported.  I'd like to steer clear of using `WeakSet` for now because of poor browser support.

src/bindings/html/view.js
stasm: With the current code, `get` and `has` are equivalent.  I don't expect us to set anything else than `true`, so I can go with `has`.

PR #87, opened 2015-10-01
Bug 1210719 - Create a bridge-enabled version of L20n for NGA

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1210719


src/runtime/web/remote.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+
+import { fetch } from './io';
+import { Remote } from '../../bindings/html/remote';
+
+export class WebRemote extends Remote {
+  constructor(requestedLangs) {
+    super(requestedLangs);
```
stasm: I move the `Remote` back into the bindings, and I'd like to extend it in the runtime.  A `remote` needs two runtime-specific things:  a `fetch` method and a `broadcast` method.  The implementation depends on the client-service architecture.

I'd like to extend the base `Remote` here, but Gecko doesn't support `super` at all (https://bugzil.la/1066239).  @zbraniecki any suggestions how to handle this?  I could re-write `Remote` to use the es5 `prototype` way of defining constructors and methods and then `Remote.call(this, requestedLangs)` here.  What do you think?

stasm:
Transpiled with Babel, l20n-client weighs in at 15kB.  If we moved qps entirely into the service, we could go down to 12kB.  But that's for another bug/pr, because we need qps for `document.l10n.qps` in some form.

PR #89, opened 2015-10-05
Bug 1211561 - Destroy contexts when view is disposed of

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1211561

This reverts commit e12460a9da165d43b4e4750fd1c4abd654d9c72b:
"Flatten the cache and remove _resMap"

Conflicts:
	src/lib/context.js
	src/lib/env.js
	tests/lib/env/resCache.js


src/lib/env.js
```diff
@@ -55,29 +67,44 @@ export class Env {
...
+    cache.set(id, resource);
+
+    return resource;
+  }
+}
+
+function deleteIfOrphan(cache, lists, resId) {
+  let isNeeded = false;
+  lists.forEach(
+    resIds => isNeeded = resIds.has(resId) ? true : isNeeded);
```
zbraniecki: It would be possibly nice to stop looping on first `true`. Is there `Map` equivalent of `Array.some`?

stasm: Unfortunately not.  I'm not a fan of `forEach` either and I haven't figured out how to use `Array.prototype.some.call` yet.

stasm: I could perhaps use `Array.from(…).some(…)`.  @zbraniecki what's your take?

stasm:
```javascript
const isNeeded = Array.from(lists).some(
  ([ctx, resIds]) => resIds.has(resId));
```

zbraniecki: nah, leave it for now. Destructing will be a rare operation and the number of contexts probably below 10, so no need to over-optimize it for now imo.

stasm: I agree it's going to be rare. Can I still use `some`? I think it tells a better story about this snippet's purpose.

zbraniecki: sure, although it defies the purpose. The only value of some is if it allows you not to cycle through the whole list even once.
With your approach you loop twice in the conservative scenario.

PR #92, opened 2015-10-13
Bug 1214135 - Put parsers into their own modules

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1214135


stasm:
We might end up doing this for Gecko, but it will be done in a different manner than in this PR.  Closing.

PR #98, opened 2015-10-28
Bug 1219208 - Always set html.langs. r=gandalf

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1219208

### DIFF
--- a/src/bindings/html/view.js
+++ b/src/bindings/html/view.js
@@ -96,7 +96,7 @@ export function translateDocument(view, langs) {
 
   if (readiness.has(html)) {
     return translateFragment(view, langs, html).then(
-      () => setDOMAttrsAndEmit(html, langs));
+      () => setAllAndEmit(html, langs));
   }
 
   const translated =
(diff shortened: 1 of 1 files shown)


src/bindings/html/view.js
```diff
@@ -104,23 +104,30 @@ export function translateDocument(view, langs) {
...
+    setLangs(html, langs);
+    readiness.set(html, true);
+  });
 }
 
-function setDOMAttrsAndEmit(html, langs) {
-  setDOMAttrs(html, langs);
+function setLangs(html, langs) {
+  const codes = langs.map(lang => lang.code);
+  html.setAttribute('langs', codes.join(' '));
```
zbraniecki: I'm still not sure if it should be space or comma separated list. Our meta use comma and we did this for a reason.

zbraniecki: but we can iron it out later, I don't expect this to be used by anyone but us for quite a while

PR #99, opened 2015-10-31
Bug 1220396 - Add browserify source transforms

askmatey:
You can configure transforms to be automatically applied when a module is loaded in a package's `browserify.transform` field. 

See: https://github.com/substack/node-browserify#browserifytransform

Example usage (*via* `node_modules`):

```js
import l20n from 'l20n/dist/bundle/web/l20n.js';
```

Error message
```sh
SyntaxError: let is a reserved identifier
```

----------

:white_check_mark: Tested successfully with [browserify](https://github.com/substack/node-browserify) `11.2.0`.


askmatey:
https://bugzilla.mozilla.org/show_bug.cgi?id=1220396

stasm:
@askmatey wouldn't it be better to import/require the es5 version of l20n.js instead?

    import l20n from 'l20n/dist/compat/web/l20n.js';

askmatey:
I was able to transform the source by fetching the package with bower into my root directory.

PR #100, opened 2015-11-05
Bug 1221821 - Enable support for evergreen browsers

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1221821

### DIFF
--- a/src/bindings/html/shims.js
+++ b/src/bindings/html/shims.js
@@ -27,3 +27,8 @@ export function getDirection(code) {
   return ['ar', 'he', 'fa', 'ps', 'ur'].indexOf(tag) >= 0 ?
     'rtl' : 'ltr';
 }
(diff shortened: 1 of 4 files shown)


src/bindings/html/shims.js
```diff
@@ -27,3 +27,21 @@ export function getDirection(code) {
...
     'rtl' : 'ltr';
 }
+
+// Opera and Safari don't support it yet
+if (navigator.languages === undefined) {
+  navigator.languages = [navigator.language];
+}
+
+// Safari 9 and iOS 9 does not support Intl
+if (window.Intl === undefined) {
```
stasm: Hmm, should this go somewhere into src/lib/ ?  AFAIR, `Intl` is used in src/lib/context.

PR #101, opened 2015-11-10
Bug 1222055 - Error on unresolvable hash value

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1222055

### DIFF
--- a/src/lib/format/l20n/ast/parser.js
+++ b/src/lib/format/l20n/ast/parser.js
@@ -75,7 +75,8 @@ class ParseContext {
     }
 
     const ch = this._source.charAt(this._index);
-    const value = this.getValue(ch, index === undefined);
+    const hasIndex = index !== undefined;
+    const value = this.getValue(ch, hasIndex, hasIndex);
     let attrs;
 
     if (value === null) {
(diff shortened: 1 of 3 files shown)


src/lib/format/l20n/ast/parser.js
```diff
@@ -75,7 +75,10 @@ class ParseContext {
     }
 
     const ch = this._source.charAt(this._index);
-    const value = this.getValue(ch, index === undefined);
+    /*jshint -W018 */
+    const hasIndex = !(index === undefined);
+    const isRequired = hasIndex;
```
stasm: Is this about the value being required?  If so, can you call this `valueRequired`?

src/lib/format/l20n/ast/parser.js
```diff
@@ -75,7 +75,10 @@ class ParseContext {
     }
 
     const ch = this._source.charAt(this._index);
-    const value = this.getValue(ch, index === undefined);
+    /*jshint -W018 */
+    const hasIndex = !(index === undefined);
```
stasm: Why not `!==`?

src/lib/format/l20n/ast/parser.js
```diff
@@ -101,16 +104,16 @@ class ParseContext {
     return entity;
   }
 
-  getValue(ch = this._source[this._index], optional = false) {
+  getValue(ch = this._source[this._index], index = false, required = true) {
```
stasm: Do you actually need the third argument?  Could you reuse `hasIndex` for this check?  Also, perhaps keep the names of the args consistent with the names of variables in the caller code?

zbraniecki: I can't fully reuse it because there's a semantic difference between the value has an index and the value is required, but I can reuse it where `hasIndex` makes it required.

stasm: OK, I see, thanks for the explanation.  One additional nit:  perhaps rename the arguments to `hasIndex` and `isRequired`?

PR #102, opened 2015-11-12
Bug 1221821 - Enable support for evergreen browsers

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1221821


zbraniecki:
slightly different approach

PR #103, opened 2015-12-08
Bug 1228348 - Allow l20n-service to run in a worker context

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1228348

### DIFF
--- a/src/bindings/html/langs.js
+++ b/src/bindings/html/langs.js
@@ -3,59 +3,8 @@
 import { prioritizeLocales } from '../../lib/intl';
 import { pseudo } from '../../lib/pseudo';
 
-export function getMeta(head) {
-  let availableLangs = Object.create(null);
-  let defaultLang = null;
-  let appVersion = null;
-
-  // XXX take last found instead of first?
-  const metas = Array.from(head.querySelectorAll(
-    'meta[name="availableLanguages"],' +
-    'meta[name="defaultLanguage"],' +
-    'meta[name="appVersion"]'));
-  for (let meta of metas) {
-    const name = meta.getAttribute('name');
-    const content = meta.getAttribute('content').trim();
-    switch (name) {
-      case 'availableLanguages':
-        availableLangs = getLangRevisionMap(
-          availableLangs, content);
-        break;
-      case 'defaultLanguage':
-        const [lang, rev] = getLangRevisionTuple(content);
-        defaultLang = lang;
-        if (!(lang in availableLangs)) {
-          availableLangs[lang] = rev;
-        }
-        break;
-      case 'appVersion':
-        appVersion = content;
-    }
-  }
-
-  return {
-    defaultLang,
-    availableLangs,
-    appVersion
-  };
-}
-
-function getLangRevisionMap(seq, str) {
-  return str.split(',').reduce((seq, cur) => {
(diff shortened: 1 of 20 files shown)


src/bindings/html/view.js
```diff
@@ -24,23 +24,27 @@ export class View {
...
 
     const observer = new MutationObserver(onMutations.bind(this));
     this._observe = () => observer.observe(doc, observerConfig);
     this._disconnect = () => observer.disconnect();
 
-    const translateView = langs => translateDocument(this, langs);
-    client.on('translateDocument', translateView);
-    this.ready = this._interactive.then(
-      client => client.method('resolvedLanguages')).then(
-      translateView);
```
stasm: Removing this double communication with the service (first `this._interactive` calls `registerView`, then this code calls `resolvedLanguages`) is what resulted in the performance win.

src/bindings/html/langs.js
```diff
@@ -68,11 +17,7 @@ export function negotiateLanguages(
     src: getLangSource(appVersion, availableLangs, additionalLangs, code),
   }));
 
-  if (!arrEqual(prevLangs, newLangs)) {
-    fn(langs);
-  }
-
-  return langs;
+  return [langs, !arrEqual(prevLangs, newLangs)];
```
zbraniecki: since you're passing this around, it would be more readable if it was an object rather than an array.

src/lib/env.js
```diff
@@ -85,11 +87,11 @@ export class Env {
     };
 
     const langToFetch = lang.src === 'pseudo' ?
-      { code: this.defaultLang, src: 'app' } :
+      { code: 'en-US', src: 'app' } :
```
zbraniecki: ugh, do you really need the lang code to be hardcoded that deep in the codebase? I'd prefer to set the source language during Env initialization.

src/bindings/html/langs.js
stasm: sure, will do

src/lib/env.js
```diff
@@ -12,12 +12,13 @@ const parsers = {
...
 export class Env {
-  constructor(defaultLang, fetchResource) {
-    this.defaultLang = defaultLang;
+  constructor(fetchResource) {
     this.fetchResource = fetchResource;
 
     this._resLists = new Map();
     this._resCache = new Map();
+    // XXX remove when sw land
+    this._appVersions = new WeakMap();
```
stasm: @zbraniecki I'm not sure if this should be defined here, or maybe I could store the `appVersion` on the `langs` objects, where we already have the language code, direction and source (`app` or `extra` for langpacks).  wdyt?

src/lib/env.js
stasm: I'd like to stop passing the default language to the `Env` constructor because in the future, we should just take the last language in the `langs` array.  This is why I don't mind hardcoding this here for the time being.

src/bindings/html/view.js
zbraniecki: :thumbsup:

src/lib/env.js
zbraniecki: I like the idea oh having it on `langs`.

stasm: Yeah, me too.  I'll change it, thanks!

PR #105, opened 2015-12-09
Bug 1231217 - Create API for observing new roots. r=gandalf

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1231217

[Review on Reviewable](https://reviewable.io/reviews/l20n/l20n.js/105)

### DIFF
--- a/src/bindings/html/observer.js
+++ b/src/bindings/html/observer.js
@@ -1,0 +1,56 @@
+'use strict';
+
+import { translateFragment } from './dom';
+
+const observerConfig = {
+  attributes: true,
+  characterData: false,
+  childList: true,
+  subtree: true,
+  attributeFilter: ['data-l10n-id', 'data-l10n-args']
+};
+
+const observers = new WeakMap();
+
+export function initMutationObserver(view, onMutations) {
+  observers.set(view, {
+    roots: new Set(),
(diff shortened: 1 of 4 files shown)


src/bindings/html/observer.js
```diff
@@ -0,0 +1,56 @@
...
+  }
+}
+
+export function observeRoot(view, root) {
+  const obs = observers.get(view);
+  obs.roots.add(root);
+  obs.observer.observe(root, observerConfig);
+}
+
+export function disconnectRoot(view, root) {
```
zbraniecki: can you merge `disconnect` and `disconnectRoot`? Also, you probably want the guard on `observers.get` result anyway.

src/bindings/html/view.js
```diff
@@ -3,22 +3,16 @@
...
-const observerConfig = {
-  attributes: true,
-  characterData: false,
-  childList: true,
-  subtree: true,
-  attributeFilter: ['data-l10n-id', 'data-l10n-args']
-};
-
-const readiness = new WeakMap();
+const priv = new WeakMap();
```
zbraniecki: Can we get a better name here? `priv` feels in the same universe as `data`, `meta`, and `thing`. :)

PR #107, opened 2015-12-12
Bug 1231418 - Make Contexts immutable

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1231418

### DIFF
--- a/src/bindings/html/remote.js
+++ b/src/bindings/html/remote.js
@@ -9,44 +9,36 @@ export class Remote {
     this.broadcast = broadcast;
     this.env = new Env(fetchResource);
     this.ctxs = new Map();
-    this.langs = new Map();
   }
 
   registerView(view, resources, meta, additionalLangs, requestedLangs) {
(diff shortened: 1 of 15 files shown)


src/bindings/html/remote.js
```diff
@@ -9,44 +9,36 @@ export class Remote {
...
-  resolvedLanguages(view) {
-    return this.langs.get(view);
+    return this.ctxs.get(view).formatValues(...keys);
   }
 
   changeLanguages(view, meta, additionalLangs, requestedLangs) {
-    const prevLangs = this.langs.get(view) || [];
-    const { langs, haveChanged } = negotiateLanguages(
+    const oldCtx = this.ctxs.get(view);
+    const prevLangs = oldCtx._langs || [];
```
zbraniecki: I think you said that there's no scenario in which `_langs` is falsy

stasm: Correct, I must have missed that in a rebase. Thanks!

PR #109, opened 2016-01-04
Bug 1236367 - Clean up the codebase for eslint

zbraniecki:
https://bugzilla.mozilla.org/show_bug.cgi?id=1236367

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
       "dot-location": [2, "property"],
       "no-implied-eval": 2,
       "no-loop-func": 2,
-      "no-magic-numbers": 1,
+      "no-magic-numbers": [1, {"ignore": [-1, 0, 1, 2]}],
       "no-useless-call": 1,
       "no-useless-concat": 1,
       "no-delete-var": 2,
@@ -134,7 +134,7 @@
       "prefer-const": 1,
       "prefer-spread": 2,
       "strict": [2, "global"],
-      "indent": [1, 2]
+      "indent": [1, 2, {"SwitchCase": 1}]
     }
   }
 }
--- a/src/runtime/web/io.js
+++ b/src/runtime/web/io.js
@@ -1,9 +1,9 @@
-'use strict';
-
 import { L10nError } from '../../lib/errors';
 
+const HTTP_STATUS_CODE_OK = 200;
+
 function load(type, url) {
-  return new Promise(function(resolve, reject) {
+  return new Promise((resolve, reject) => {
     const xhr = new XMLHttpRequest();
 
     if (xhr.overrideMimeType) {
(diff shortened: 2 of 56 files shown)


package.json
```diff
@@ -94,7 +94,7 @@
       "dot-location": [2, "property"],
       "no-implied-eval": 2,
       "no-loop-func": 2,
-      "no-magic-numbers": 1,
+      "no-magic-numbers": [1, {"ignore": [-1, 0, 1, 2]}],
```
zbraniecki: I feel like it doesn't make sense to always document/explain all those numbers -
 - `-1` usually is for negative indexOf
 - 0, 1, 2 are usually for first character in the string, or tuple elements.

package.json
```diff
@@ -134,7 +134,7 @@
       "prefer-const": 1,
       "prefer-spread": 2,
       "strict": [2, "global"],
-      "indent": [1, 2]
+      "indent": [1, 2, {"SwitchCase": 1}]
```
zbraniecki: `SwitchCase` means that we do:
```
switch () {
  case 'foo':
    return x;
}
```

instead of:
```
switch () {
  case 'foo':
  return x;
}
```

src/runtime/web/io.js
```diff
@@ -16,10 +16,9 @@ function load(type, url) {
       xhr.responseType = 'json';
     }
 
-    xhr.addEventListener('load', function io_onload(e) {
-      if (e.target.status === 200 || e.target.status === 0) {
-        // Sinon.JS's FakeXHR doesn't have the response property
-        resolve(e.target.response || e.target.responseText);
+    xhr.addEventListener('load', e => {
+      if (e.target.status === HTTP_STATUS_CODE_OK) {
```
zbraniecki: Seems to me that we don't handle build time with this code anymore so we don't need to support the non-standard HTTP status `0`.

src/runtime/web/io.js
```diff
@@ -16,10 +16,9 @@ function load(type, url) {
...
+        resolve(e.target.response);
```
zbraniecki: Sinon added `response` - https://github.com/sinonjs/sinon/commit/2ff8229e0bce65f20815f11927736559834cafb5

package.json
stasm: I feel like for tuples we should rather use destructuring than indexes. For strings, magic numbers are fine, so this looks okay to me.

package.json
stasm: :+1:

src/runtime/web/io.js
stasm: Wasn't `0` used also when accessing `l20n.js` via the `file://` protocol in the browser?

src/runtime/web/io.js
stasm: Yay!

src/runtime/web/io.js
zbraniecki: hmm, maybe it was, I'll readd it for now.

PR #110, opened 2016-02-17
Bug 1245500 - Add polyfills for new String methods to deps

stasm:
Add the following polyfills to runtime dependencies:

  string.prototype.startswith
  string.prototype.endswith

https://bugzilla.mozilla.org/show_bug.cgi?id=1245500


zbraniecki:
Why does this patch change eslint settings?

stasm:
Because I used npm to edit `package.json` and it reformatted the JSON.

PR #111, opened 2016-02-22
Allow class attribute for tags within translations.

poscar:
This adds 'class' attribute to be used in tags within translations. 

One useful usecase allowed by this change is to be able to specify font awesome icons within a string. This is done by just inserting <i class="fa fa-paperclip"></i> within the string in the l20n file.

### DIFF
--- a/src/bindings/html/overlay.js
+++ b/src/bindings/html/overlay.js
@@ -11,7 +11,8 @@ const allowed = {
     'mark', 'ruby', 'rt', 'rp', 'bdi', 'bdo', 'span', 'br', 'wbr'
   ],
   attributes: {
-    global: [ 'title', 'aria-label', 'aria-valuetext', 'aria-moz-hint' ],
+    global: [ 'title', 'aria-label', 'aria-valuetext', 'aria-moz-hint',
+      'class' ],
     a: [ 'download' ],
     area: [ 'download', 'alt' ],
     // value is special-cased in isAttrAllowed
@@ -192,4 +193,3 @@ function camelCaseToDashed(string) {
     })
     .replace(/^-/, '');
 }
-


poscar:
Do I need to file a bug to merge this?

stasm:
Sorry for the wait.  This is an interesting usecase which I'd like to discuss further.  I can't speak to the Font Awesome case but I know that Google's Material Icons have switched to using ligatures instead of class names which is better for accessibility.  Are there any plans to do the same for Font Awesome?

I'd be reluctant to add the `class` attribute to the whitelist right now.  I'd like to better understand any implications this might have for security.  With access to classes, a malicious translation could easily modify the layout of the UI.

poscar:
As far as I know there are no plans to implement ligatures in Font Awesome. It would nice if they did.

I am not sure what the security implications of class attributes would be. It will just affect the style of child elements to the localized element. I do not think there's risk of code execution or anything like that. The worse case that I can think of is if a malicious localization is able to fool the user to enter certain information by making the page look a certain way.  But then again, if localization files have been compromised, there's a bigger security issue at play.  I think forbidding class attributes does not really add any meaningful security benefits. What are your thoughts?

stasm:
@mozfreddyb, any chance you could chime in here?  We're wondering whether allowing localizers to edit the `class` attribute in HTML could be allowed.  As @poscar was right to point out, at some point we should assume that a compromised localization can do much more harm by just being wrong than by modifying certain DOM attributes.

cc @zbraniecki

mozfreddyb:
I was hoping that a compromised localization can only do *limited harm*
I'm sure we defend against this attack when it comes to script injections.

But the question is where do you want to draw the line?
Sometimes a `id="attributes"` can be enough to clobber (i.e., overwrite) DOM properties on HTML nodes that will confuse JavaScript (e.g. `parentNode.attributes` in JavaScript now points to the child node with the `id="attributes"`)

It's very hard to make assumptions on every web page that might ever use l20n.js.
Simple `class` attributes can have funky side-effects with certain third party libraries. You can enable AngularJS for a domtree by giving a parent element the `ng-app` class (this stopped working in more recent versions of Angular though)

poscar:
@mozfreddyb Completely agree that a compromised localization should cause limited harm. But there are attributes that are useful when specifying markup within a translated string. These are things like specifying which part of the translation should have italics, bold, or a specific style. Given this, being able to specify classes is a valuable feature. I also think it should be up to the users to know what they can and can't use in translations if they are using a third party library.

Id and other attributes that have more serious side effects should still not be allowed, but I believe the benefits of allowing class attributes outweigh the risks. What do you think?

Thanks!

zbraniecki:
I'm in to allow the `class` attribute

poscar:
@stasm should I switching this PR to master as well, or is v3 still a live branch?

stasm:
Yes please, let's rebase this on top of master.  v3 is a release branch -- we merge master to it when it's stable.

PR #114, opened 2016-03-01
FTL parser

stasm:
The purpose of this PR is to give feedback to @zbraniecki.


src/lib/format/ftl/ast/ast.js
```diff
@@ -0,0 +1,136 @@
...
+
+class Trait extends Node {
+  constructor(id, value) {
+    super();
+    this.id = id;
+    this.value = value;
+  }
+}
+
+class Variant extends Node {
```
stasm: Variants and Traits should be the same:  Members.  `entry.traits` and `placeable.variants` should both be Arrays of Members.

src/lib/format/ftl/ast/ast.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+class Placeable extends Node {
+  constructor(content = null) {
+    super();
+    this.content = content;
+  }
+}
+
+class SelectExpression extends Node {
```
stasm: I think we can merge Pleaceable and SelectExpression into one Node, like this:

```
{ brand-name }
```

src/lib/format/ftl/ast/ast.js
```diff
@@ -0,0 +1,136 @@
...
+  }
+}
+
+class Value extends Node {
+  constructor() {
+    super();
+  }
+}
+
+class String extends Value {
```
stasm: String and Value can be merged into one Node.

src/lib/format/ftl/ast/ast.js
```diff
@@ -0,0 +1,136 @@
...
+  constructor(source, content) {
+    super();
+    this.source = source;
+    this.content = content;
```
stasm: Perhaps `elements` would be a better name?

src/lib/format/ftl/ast/ast.js
stasm: Or, event simpler:  the value should be defined on the Entry as an empty array.

stasm: And then each Pleaceable would have a `source` property.

src/lib/format/ftl/ast/ast.js
```diff
@@ -0,0 +1,120 @@
...
+  constructor(name) {
+    super();
+    this.name = name;
+  }
+}
+
+class String extends Node {
+  constructor(source, elements) {
+    super();
+    this.source = source;
```
stasm: Could we move `source` from String to Placeable?  I realized that FTL is an opportunity to finally do per-placeable fallback to source, what with all selection logic inside of each placeable.

stasm:
@zbraniecki I've been thinking some more about strings and values.  First, I take back what I said about `Entity.value` being an array.  I think it makes sense to have an intermediate AST node for complex strings.

The word _string_ might be confusing here.  What I mean is a value _pattern_  which can consist of text elements and placeables.  OTOH, a _string literal_ could only consist of a single text element.  Here's a simplified PEG syntax:

```
Pattern
    = Element+
    / '"' Element+ '"'
    / ('\n|' Element+)+

Element
    = PlaceableElement
    / TextElement

PlaceableElement
    = '{' Expression ( '->' Member+)* '}'

StringLiteral
    = '"' TextElement '"'

TextElement
    = TextCharacter*

TextCharacter
    = [^"{]
    / '\\"'
    / '\\{'
```

I'd start with `Pattern` being only allowed in value positions (`Entity.value` and `Member.value`), and `StringLiteral` being allowed inside of placables and kwargs.  Also, `StringLiterals` wouldn't be allowed to use the pipe syntax for now.  Granted this would make it impossible to do things like https://github.com/stasm/l20n-syntax-experiments/blob/0434d1f15fb1ab9a76f8e94771f2c986b1caa3f7/ftl/05-complex.txt#L7-L11 but that's a very complex example and requires `Intl.ListFormat` which isn't implemented yet anyways.

Thoughts?

PR #115, opened 2016-03-23
Fixed square boxes in Microsoft browsers, including Edge.

poscar:
The FSI and PDI characters are displayed as square boxes in Microsoft browsers (see attached screenshot). This pull request fixes that issue.

I know it removes the FSI and PDI markers, but I believe disabling them is better than incorrectly displaying square boxes for any replaced strings that result in the subPlaceable function being called.

screen-shot-2016-03-22-at-8 05 49-pm


zbraniecki:
I think that we should make it configurable if anything. Removing functionality that is used by products relying on l20n (that don't target Edge), is not a solution.

poscar:
@zbraniecki What would be the right way to make this configurable?  Separate dist?

stasm:
@poscar Can you please create your PR against master rather than 3.x?

stasm:
We could use `<bdi>` instead of the Unicode marks.

poscar:
@stasm, Ok. I'll try out <bdi> and make the pull request against master.

PR #116, opened 2016-03-23
Switching unicode markers to bdi tags to fix MS browsers showing box characters in text.

poscar:
The FSI and PDI characters are displayed as square boxes in Microsoft browsers (see attached screenshot). This pull request fixes that issue by using bdi tags instead.

screen-shot-2016-03-22-at-8 05 49-pm

### DIFF
--- a/src/lib/resolver.js
+++ b/src/lib/resolver.js
@@ -4,8 +4,8 @@ const KNOWN_MACROS = ['plural'];
 const MAX_PLACEABLE_LENGTH = 2500;
 
 // Unicode bidi isolation characters
-const FSI = '\u2068';
-const PDI = '\u2069';
+const FSI = '<bdi>';
+const PDI = '</bdi>';
 
 const resolutionChain = new WeakSet();
 
--- a/tests/lib/util.js
+++ b/tests/lib/util.js
@@ -1,8 +1,8 @@
 'use strict';
 
 // Unicode bidi isolation characters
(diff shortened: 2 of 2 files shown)


Pike:
Sadly https://connect.microsoft.com/IE/Feedback/Details/2457310 only has the stock @MicrosoftEdge response.

poscar:
@stasm Could we merge this one as well?

zbraniecki:
Ok, so Microsoft is investigating - https://developer.microsoft.com/microsoft-edge/platform/issues/6831905/. I can reproduce this bug in Virtualbox. I'm not sure if it's a virtualbox only or does it also happen in full environment?

I'd like to keep this issue open, but I'm not sure if we should do what the patch suggests. In particular, this would not work in non-HTML environments (Node, JS etc.) so we'd need per-bindings solution.
I'd rather see MS resolve it :)

poscar:
@zbraniecki Cool, it'll be great if they fix it. Only issue then will be that older browsers will probably not get the fix.

zbraniecki:
>  Only issue then will be that older browsers will probably not get the fix.

Edge is on the train release so I expect them to deliver an update to their users with the fix before we get l20n 4.0 released.

tomekstach:
@zbraniecki Unfortunately, Edge still has the same problem. Do you have any idea how prepare a workaround for Edge?

PR #117, opened 2016-03-23
Allow class attribute for tags within translations.

poscar:
This adds 'class' attribute to be used in tags within translations.

One useful usecase allowed by this change is to be able to specify font awesome icons within a string. This is done by just inserting within the string in the l20n file.

This pull request is against master, there was previous discussion to allow this feature in pull request #111


poscar:
@stasm Any chance of merging this one?

zbraniecki:
I'm still torn about this proposal. It opens up a new attack vector where a localization file may include a localizable element into a selector that gets bind to an action, which is scary.

For example, imagine code that does `$(".remove-all").click(doRemoveEverything);` and then a localizer assigns `remove-all` class to an element with label `Do Nothing`.

That's scary.

PR #119, opened 2016-05-07
Use generators and flatMap to handle errors and context

stasm:
It has always bugged me that we need to pass the `ctx` and `args` into all functions of the resolver.  And returning `[value, errors]` tuples to 1) allow for better and more graceful fallback and 2) avoid `try…catch` meant that composing functions was a nightmare.

I'd like to suggest a pattern which abstracts the extra output (errors) and extra input (context data) away from the actual code of the resolver.  Thanks to generators a wrapper function can handle these side effects (right when `yield` is called and the actual function is suspended) and—once ready—it can inject just the values that the resolver cares about back into its code (with the iterator's `next()` method).

The code becomes much clearer and cleaner IMO, and the tedious task of handling errors is no more.  There's a performance penalty related to using generators (which I hope one day will be optimized in the engine).  Compared to my recent changes which brought the benchmark down to 10 ms on my machine using jsshell, we're back to 15-16 ms, which is what I started with this morning.

compared to current master:

    mean:   16104.7 (+60%)
    stdev:  558.54
    sample: 30

compared to yesterday's master:

    mean:   16044.4 (+3%)
    stdev:  760.71
    sample: 30

You could say I had won these 5 ms to be able to suggest this change :)

We can probably look for other optimizations (which will be easier now that the code is easier to modify) and try to go down to 10 ms again.  @zbraniecki, what do you think?

### DIFF
--- a/src/lib/builtins.js
+++ b/src/lib/builtins.js
@@ -10,6 +10,12 @@ class FTLBase {
   }
 }
 
+export class FTLNone extends FTLBase {
+  toString() {
+    return this.value || '???';
+  }
+}
+
 export class FTLNumber extends FTLBase {
   constructor(value, opts) {
     super(parseFloat(value), opts);
--- a/src/lib/builtins.js
+++ b/src/lib/builtins.js
@@ -20,9 +20,9 @@ export class FTLNumber extends FTLBase {
   constructor(value, opts) {
     super(parseFloat(value), opts);
   }
-  toString(rc) {
-    const nf = rc.ctx._memoizeIntlObject(
-      L20nIntl.NumberFormat, rc.lang, this.opts
+  toString(ctx, lang) {
+    const nf = ctx._memoizeIntlObject(
+      L20nIntl.NumberFormat, lang, this.opts
     );
     return nf.format(this.value);
   }
@@ -32,9 +32,9 @@ export class FTLDateTime extends FTLBase {
   constructor(value, opts) {
     super(new Date(value), opts);
   }
-  toString(rc) {
-    const dtf = rc.ctx._memoizeIntlObject(
-      L20nIntl.DateTimeFormat, rc.lang, this.opts
+  toString(ctx, lang) {
+    const dtf = ctx._memoizeIntlObject(
+      L20nIntl.DateTimeFormat, lang, this.opts
     );
     return dtf.format(this.value);
   }
@@ -45,7 +45,7 @@ export class FTLKeyword extends FTLBase {
     const { name, namespace } = this.value;
     return namespace ? `${namespace}:${name}` : name;
   }
-  match(rc, other) {
+  match(ctx, lang, other) {
     const { name, namespace } = this.value;
     if (other instanceof FTLKeyword) {
       return name === other.value.name && namespace === other.value.namespace;
@@ -54,8 +54,8 @@ export class FTLKeyword extends FTLBase {
     } else if (typeof other === 'string') {
       return name === other;
     } else if (other instanceof FTLNumber) {
-      const pr = rc.ctx._memoizeIntlObject(
-        L20nIntl.PluralRules, rc.lang, other.opts
+      const pr = ctx._memoizeIntlObject(
+        L20nIntl.PluralRules, lang, other.opts
       );
       return name === pr.select(other.valueOf());
     }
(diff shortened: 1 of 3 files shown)


stasm:
I reviewed the code and relaized that in some places yielding wasn't necessary because no side effect could occur.  I was thus able to improve the performance significantly.  The new resolver is only 2-3 ms slower than master's:

compared to master: 

    mean:   12825.47 (+27%)
    stdev:  427.62
    sample: 30

stasm:
New numbers after I landed 8941e36c484f6e228cb4dccd95c70455fac45335:

compared to master (jsshell):

    mean:   10638.47 (+28%)
    stdev:  209.87
    sample: 30

compared to master (node):

    mean:   10742 (+22%)
    stdev:  1086.52
    sample: 30

zbraniecki:
I didn't review the patch yet, so everything below are some high level brainstorming thoughts.

I looked at the perf numbers. According to what I see we have:

1) Node
```
format: 
  mean:   12083.07 (+21%)
  stdev:  591.1
  sample: 30

```

2) JSShell
```
format: 
  mean:   12116.77 (+4%)
  stdev:  453.71
  sample: 30
```

which is acceptable.

I'll look further into the patch, and I'd love to aim for two things:

1) Perf win. I understand that you assume generators are not optimized, but can you help me find any article/doc/discussion that would indicate that it'll ever be faster? Or some bug number in which people are claiming that they're going to get significant speedups?
Generators, as far as I understand are supposed to give us a *significant* perf (and mem?) win, and so far they seem to cost more despite doing less, which feels like a bad technology to switch to :/

2) If we can't get a perf win, minimize the perf cost. Using Intl object memoization to mask the perf cost here is imho unfair, because all you're saying is that we could get the win from memoization on top of the current perf.

Resolver is not a critically dynamic part of L20n at the moment and I don't feel like we need to iterate a lot in terms of features there. So quite honestly, to make l20n appealing, I believe we need perf/mem win much, much more than we need resolver code that is easier to work with. (also, using generators probably moves us further away from algo's that we will be able to standardize).

I'll look into the code itself and try to wrap my head around it.

stasm:
> I didn't review the patch yet, so everything below are some high level brainstorming thoughts.

Thanks for looking into this so quickly!

> I looked at the perf numbers. According to what I see we have: (snip)

Your numbers seem to be more favorable than mine which is great.  Especially on jsshell :)

1) Perf win. I understand that you assume generators are not optimized, but can you help me find any article/doc/discussion that would indicate that it'll ever be faster? Or some bug number in which people are claiming that they're going to get significant speedups?

I expect that with time and more usage there will more incentive to optimize generators.  Today, generators aren't optimized by JIT compilers. Baseline compilers can go a long way at speeding them up, though.  My primary source is the following bug:

https://bugzilla.mozilla.org/show_bug.cgi?id=903457

From there, you can find links to Andy's blog:

https://wingolog.org/archives/2013/06/11/ecmascript-generators-from-a-performance-perspective
https://wingolog.org/archives/2014/11/14/generators-in-firefox-now-twenty-two-times-faster

> Generators, as far as I understand are supposed to give us a significant perf (and mem?) win, and so far they seem to cost more despite doing less, which feels like a bad technology to switch to :/

No, this patch and this approach has little to do with winning performance.   It has everything to do with code which is easier to manage and reason about.  This means fixing bugs, finding optimization opportunities and extending the feature set (or reducing it) will be easier going forward.

Ideally, this patch would be on par with the current performance, but I'm willing to accept a slight regression if it means better overall quality.

2) If we can't get a perf win, minimize the perf cost. Using Intl object memoization to mask the perf cost here is imho unfair, because all you're saying is that we could get the win from memoization on top of the current perf.

I landed Intl memoization on master first so that the comparison here *is* fair.  My point was that we can optimize in other places too, if we need and want.  I've been already able to shave off 4 ms earlier today which hopefully demonstrates my commitment to the matter :)

> Resolver is not a critically dynamic part of L20n at the moment and I don't feel like we need to iterate a lot in terms of features there. So quite honestly, to make l20n appealing, I believe we need perf/mem win much, much more than we need resolver code that is easier to work with. (also, using generators probably moves us further away from algo's that we will be able to standardize).

We don't know yet what it is that we want to standardize so for now I'd like to have code that's easy to extend if necessary.  Resolver happens to be the most complex part of the stack because of how resilient to errors it needs to be and how many possibilities there are.  Having code which is easier to reason about is hugely helpful and would make our work easier.

Let me know if you have more questions!

zbraniecki:
r+ :)
Land at will!

stasm:
Thanks!

PR #120, opened 2016-05-10
New low-level API

stasm:
This is using Message and Bundle classes which are stubs for the future Intl.* formatters.

Also removed: env.js, pseudo.js.  I'm in the process of removing client/service right now.  I'll do this on this branch. I've removed all caching on purpose for now.

This is one huge WIP.  For now the following node script works:

```javascript
'use strict';

require('babel-register')({
  plugins: ['transform-es2015-modules-commonjs']
});

const { fetchResource } = require('./src/runtime/node/io');
const { createContext } = require('./src/lib/context');

createContext(['./foo.ftl'], 'pl', fetchResource).then(ctx => {
  const about = ctx.formatValues(['about']);
  console.log(about);
}).catch(console.error);
```

Consult context.js, bundle.js and message.js to see how this works.

For this approach to work best it would be helpful for each parsed entry to have information about which language it's in.  This allows the entries to be flat but each entry can use the correct number or date formatters.

I've started working on HTML's View.  I'll finish later tonight.


stasm:
I don't know what names we'll end up with, but for now:

 - message.js is a singe Entity object with `.format()` and `.formatToObject()`,
 - bundle.js is a group of Entities.  It can be single- or multilingual, it doesn't matter.  It's up to the userland code to handle resources correctly.  Right now, our userland code merges and flattens resources, but in the future it could create one Bundle for each language to support better fallback.
  - context.js is the userland code. it has convenience methods like `.formatValues()` and handles the creation of the bundle.

stasm:
I'm really happy with where I got today.  Here's the summary of the changes:

  - I removed the Message class;  as @zbraniecki was right to note most translations are simple strings and it wasn't economical to create new objects for all of them.  Instead, the Bundle class has a `.format()` method which takes an *object* to format and the args.  I decided to pass an object because this allows the same method to be reused for formatting traits if needed.  In order to get the object to format, you first need to get it with `bundle.messages.get()`.  This API suggests that traits are used internally by messages but it also allows the userland code to query traits if needed.

  - Bundles are single-language, and only such. This makes it much easier to create Intl formatters and to reason about what's in the bundle.  They are also fully synchronous.  Resources are added with `bundle.addMessages()` and are parsed inside of the bundle.  Resource handling must be done by the userland code, or in our case, by the Context.  Userland can cache the resources too, although this isn't implemented right now.

  - Contexts are abstractions for fallback strategies.  For now, there's only a synchronous `SimpleContext` which falls back to the identifier.  We can add more strategies in the future, e.g. a context which prefetches more than one language and create a bundle for each of them.  Or a context which fetches only one language to start with and is able to asynchronously fetch more if need be.  In order for this to work the bindings code must take certain measures to only call  `ctx.formatValues()` and `ctx.formatEntities` in tail positions so that they can transparently return both sync values or promises.  An example of this is `View.formatValue()`.

  - This is almost ready to merge but I'd like to consider one more change:  I'd like to enforce a clearer separations between what we'd like to become part of `Intl` on one hand and our library on the other.  I suggest we create the following folder structure:
    - `src/intl` with parsers, bundle, resolver, builtins, shims and plurals,
    - `src/lib` with errors, contexts and in the future with: events, pseudo, env?

stasm:
An implication of the single-language Bundles is that if we use l10n merge to flat-merge translations into one file and we use the `SimpleContext`, fallback translations will use Intl formatter in the first language.  I think this is okay becasue `SimpleContext` is just that, simple, and shouldn't try to be too smart.

For a more advanced fallback strategy in the flat-merge scenario we'd need a way for the merging tool to annotate translations with the language they're in.

src/lib/bundle.js
```diff
@@ -0,0 +1,37 @@
+import FTLRuntimeParser from './format/ftl/entries/parser';
+import { format } from './resolver';
+
+export class Bundle {
```
stasm: I named this `Bundle` as in the MessageBundle, but perhaps this could called Context or MessageContext.  If we change the name then Context could be called something else to avoid confusion, but I'm out of ideas.

src/lib/context.js
```diff
@@ -1,134 +1,97 @@
...
-      formatted.attrs = Object.create(null);
-      for (let trait of entity.traits) {
-        const [attrValue] = format(this, lang, args, trait);
-        formatted.attrs[trait.key.name] = attrValue;
-      }
-    }
-
-    return formatted;
+  formatValues() {
+    throw new L10nError('Not implemented');
```
stasm: Is this a pythonism?

src/lib/context.js
```diff
@@ -1,134 +1,97 @@
+import { Bundle } from './bundle';
 import { L10nError } from './errors';
-import { format } from './resolver';
-
-const IntlObjects = new WeakMap();
 
 export class Context {
-  constructor(env, langs, resIds) {
+  constructor(langs, resIds) {
```
stasm: I think it's a bit confusing that `View` and `Context` work with `langs` which are `{ code, src }` objects, and `Bundle` uses `lang` which is a string code (`en-US`).  Any suggestions how to improve this?

stasm:
I'm imagining `Bundle.addMessages()` a bit like `CSSStyleSheet.insertRule()`.  It takes a string and the parsing is done internally.  This has a number of consequences:

  - because they use data and not URIs of resources, bundles are completely agnostic to the method of fetching translations.
  - because they use textual representation of data, bundles can also be standardized around the FTL syntax and not around the JSON parse tree.
  - we can cache fetched resources in the library (in Context, or we can bring the Env back, or use window.caches, or use a Service Worker, or chrome registry etc..) but the parsed data structure is stored inside of the bundles.  If bundles are part of Intl, could there also be a transparent caching mechanism there for parsed entries?
  - since parsing is done inside of the bundles we need to find a way to implement pseudolocalizations.  One solution is to just pseudolocalize the result of `bundle.format()`, which has the advantage of being very fast (no need to walk the AST and pseudolocalize text nodes) but would also affect external arguments.

stasm:
@zbraniecki Any thoughts on my last comment as well as on splitting the codebase into `src/lib` and `src/intl`?

src/lib/bundle.js
zbraniecki: I like `MessageContext` and we can think about renaming our `L20nContext` to something else, but I'm ok with leaving the names for now.

src/lib/context.js
zbraniecki: Not confusing for me. L20n stuff has many sources, Intl stuff only knows the locale code. :)

src/bindings/html/langs.js
```diff
@@ -1,4 +1,4 @@
-import { prioritizeLocales } from '../../lib/intl';
+import { prioritizeLocales } from '../../lib/shims';
```
zbraniecki: Not sure if we'll ever standardize `prioritizeLocales`, thus I kept it as a 'l20n intl', now a shim.

src/lib/context.js
```diff
@@ -1,134 +1,97 @@
...
+  }
 
-    if (!cache[id]) {
-      cache[id] = new ctor(code, opts);
-      IntlObjects.set(ctor, cache);
-    }
 
-    return cache[id];
+  formatValues(...keys) {
+    return this._formatKeys(keys, SimpleContext.prototype.formatValue);
```
zbraniecki: `this.prototype.formatValue` would work if any class inherits from SimpleContext, right?

zbraniecki:
r+! Looks good to me :)

zbraniecki:
I like the directory split, so go do it.

wrt. code itself, I'm not sure if I like storing strings on the Bundle. I'd probably prefer storing them on the L20nContext and using bundle just to format them, but we can iron it out next week.

src/lib/context.js
stasm: I think it has to be `this.constructor.prototype.formatValue` but yeah, good call.

src/lib/context.js
stasm: Is it okay to use the same name (lang) on both meanings?

zbraniecki: yeah

PR #122, opened 2016-05-16
Localization object

stasm:
In #120 I landed a new version of our low-level API.  In it, Contexts are abstractions for fallback strategies. The language negotiation, the fetching of resources and finally the parsing and the formatting happens in different modules.

This PR is about the next step in the evolution of the low level API.  It's based on a few long discussions with @Pike over the last three days.  The purpose is to prepare the API to handle different sources of translations, different resource loading methods and different methods of negotiating languages.  This, in turn, will make it easier to integrate L20n into Gecko.

Let me start with suggesting a few renames (as discussed with Pike):

    Intl.Bundle → Int.MessageContext
    L20n.Context → L20n.Localization
    L20n.View → L20n.HTMLLocalization, a subclass of L20n.Localization

The language negotiation is runtime-specific.  The input depends on the runtime:  it can be `navigator.languages` and `<meta name=”availableLanguages”>` in the web runtime, or the `matchOS` preference in Gecko, or the `Accept-Languages` header in server/service-worker-enabled runtime.  It also takes the list of resources for which it is supposed to negotiate the best translations.

The output of the language negotiation is a list of lazy resource bundles (i.e. the fallback chain).  Each bundle corresponds to one language+source pair.  Certain runtimes might want to optimize by prefetching the first group. Bundles have a `.fetch()` method which can be used by the Localization object to create `Intl.MessageContext`s, each for one bundle.  `MessageContext`s are responsible for parsing and formatting translations. If formatting of a translation returns errors, the `Localization` object can fallback to the next bundle from its current fallback chain.

In this PR I've simplified the directory structure in `src/bindings`.  I think that in the future we can have `html.js` and `xul.js` next to each other and there aren't enough files related to each to warrant separate directories.

I've also removed the `View` class and instead just subclasses `Localization` for HTML.  This may or may not be good, but for now it doesn't really matter b/c we only have a limited number of bindings to start with.  Once we create more it might make sense to see which parts of the logic could be abstracted better.  For now, however, I don't want to bother too much about the right split.

It's important to note that the PR assumes that different bindings and runtimes might end up with different `Localization` subclasses; some sync, other async.  The purpose of `Localization` instances is to start with a list of lazy resource bundles, create `MessageContext`s for them and fall back between them as necessary.


stasm:
@zbraniecki This is WIP but I'd love to hear your thoughts early on.  I still need to clean things up and also I'm not sure where the language negotiation piece should live and how exactly it should look like.  I know I want it to return a list of lazy bundles.  I think it should be possible to have different negotiators in different runtimes, e.g. gecko bindings will negotiatie via an API to the chrome registry,

stasm:
Alright, I don't think this is a WIP anymore.  It works and does everything I needed it to.  I'm happy with language negotiation moving into `runtime/.  This should make it easy for `runtime/gecko` to use chrome registry for negotiating languages and providing resources.

src/runtime/web/polyfill.js
```diff
@@ -0,0 +1,10 @@
+// Polyfill NodeList.prototype[Symbol.iterator] for Chrome.
+// See https://code.google.com/p/chromium/issues/detail?id=401699
+if (typeof NodeList === 'function' && !NodeList.prototype[Symbol.iterator]) {
+  NodeList.prototype[Symbol.iterator] = Array.prototype[Symbol.iterator];
+}
```
zbraniecki: this has been fixed in Chrome 51, can we remove it pls?

src/runtime/web/index.js
```diff
@@ -1,6 +1,35 @@
-import { createSimpleContext } from './api';
-import { View } from '../../bindings/html/view';
+import { prioritizeLocales } from '../../intl/index';
+import { HTMLLocalization } from '../../bindings/html';
+import { documentReady, getResourceLinks, getMeta } from './util';
```
zbraniecki: so, `getResourceLinks` and `getMeta` will be different for HTML and XUL. Before your patch we have `bindings/html/head` and we can add `bindings/xul/head`. Not sure how you want to handle that here.

src/runtime/web/index.js
```diff
@@ -1,6 +1,35 @@
...
+function createResourceBundle(resIds, lang) {
+  return {
+    lang,
+    loaded: false,
+    fetch() {
+      return this.loaded || (this.loaded = Promise.all(
+        resIds.map(id => fetchResource(id, lang))
+      ));
+    }
+  };
```
zbraniecki: why anonymous object instead of a class?

src/lib/localization.js
```diff
@@ -0,0 +1,52 @@
+import { L10nError } from './errors';
+
+export class Localization {
```
zbraniecki: I'm not sure what's the value of a class that has only private methods and doesn't give anything to `HTMLLocalization`. Maybe it should be just a bunch of functions?

src/runtime/web/index.js
stasm: No reason, I can change it to class.

src/lib/localization.js
stasm: Possibly.  I'm still considering renaming this to `FallbackStrategy` which is used by `HTMLLocalization`.  However, I think regular functions will work jus fine in the current scenario, since we only have one strategy anyways for now.  I'll change it.

src/runtime/web/polyfill.js
stasm: Yes!

src/runtime/web/index.js
stasm: The difference is that `getResourceLinks` and `getMeta` is not bindings-specific, but runtime-specific. This allows you to have HTML bindings in a Service Worker runtime for instance which won't have `getMeta` at all.  In this PR I'm making it clearer (hopefully!) that the bindings are about *how* things are localized, and runtime is about plugging the localization in and initializing it.  Language negotiation is the first step to initialize the `Localization` object and thus I thought it would be fitting to put it in `runtime/`.

stasm: So with this patch, you can have `runtime/gecko` which defines its own `getResourceLinks` (querying for `<messagebundle>`s) and doesn't have `getMeta` at all.

stasm: At the same time, the `Localization` class doesn't even know about `meta` or `links` or `messagebundles`.  All it cares for is an initialization promise which resolves to a list of lazy bundles, which is the _de facto_ current language negotiation state. This means that we can use the same `HTMLLocalization` in different runtimes, e.g. web (which requires `meta`) and gecko (which doesn't because the chrome registry returns the right resource file).

PR #123, opened 2016-05-17
Custom macros

stasm:
This PR adds custom macros to `MessageContext`.  The code is largely taken from @zbraniecki's gecko-dev branch.  Since macros can now be custom I'm also changing `BuiltinReference` to `MacroReference`.

I've used 'macro' here because that's what we've always used to refer to functions available to translations.  However, macros are historically more about expanding one string to another, or transforming one part of AST into another. This is not what is going on here.  It occurs to me that calling them simply `Function`s might be more suitable.

@zbraniecki, what's your preference?

I know you used `formatters` on your branch, but things like `TAKE()` or `DROP()` aren't formatters.  Macros/functions can make use of formatters internally.


zbraniecki:
Macros vs Functions - not sure. Excel uses Functions, right? Maybe that validates use of the word?

But I'd suggest not switching IntlObject to Formatter - PluralRules is not a formatter :) Collator is not a formatter, etc.

stasm:
17.05.2016 18:30 "Zibi Braniecki" <notifications@github.com> napisał(a):
>
> Macros vs Functions - not sure. Excel uses Functions, right? Maybe that
validates use of the word?

Yeah, I was thinking that, too. Should I change it to Function?

>
> But I'd suggest not switching IntlObject to Formatter - PluralRules is
not a formatter :) Collator is not a formatter, etc.

Ah, right you are. IntlObject it is, then!

zbraniecki:
r+

PR #124, opened 2016-05-18
Intl.MessageContext

stasm:
This is my attempt at making it easy to use `MessageContext` in different environments.  Instead of exporting the constructor (which is troublesome in Gecko's JSM), we expose it on the global `Intl` object.


stasm:
With this change we only need to do `import '../../intl/polyfill';` in the runtime and `Intl.MessageContext` becomes available everywhere.  Yay globals :)

@zbraniecki is this along the lines of what you've been thinking about?

PR #125, opened 2016-05-18
Initial Gecko landing

zbraniecki:
test PR so that I can comment inline. Won't merge.


src/bindings/dom.js
```diff
@@ -13,6 +13,7 @@ export function setAttributes(element, id, args) {
   if (args) {
     element.setAttribute('data-l10n-args', JSON.stringify(args));
   }
+  return element;
```
zbraniecki: It allows me to chain that into JS code like this: https://github.com/zbraniecki/gecko-dev/blob/2e14a3a53d6476002f3c4f79e6871376c6056b1e/toolkit/content/aboutSupport.js#L479

src/bindings/dom.js
```diff
@@ -89,7 +90,11 @@ function translateElements(view, elements) {
 function applyTranslations(view, elems, translations) {
   disconnect(view, null, true);
   for (let i = 0; i < elems.length; i++) {
-    overlayElement(elems[i], translations[i]);
+    try {
+      overlayElement(elems[i], translations[i]);
+    } catch (e) {
+      console.log(e);
+    }
```
zbraniecki: that means that if overlayElement fails, we don't break the rest of UI. Not sure if we should land it or make overlayElement more robust.

src/bindings/overlay.js
```diff
@@ -10,6 +10,7 @@ const allowed = {
   ],
   attributes: {
     global: ['title', 'aria-label', 'aria-valuetext', 'aria-moz-hint'],
+    button: ['accesskey'],
```
zbraniecki: I need this for XUL. Overall we may want to have different whitelist for XUL and HTML. I really think that bindings/html and bindings/xul directories make sense

src/bindings/overlay.js
```diff
@@ -34,7 +35,8 @@ export function overlayElement(element, translation) {
     } else {
       // start with an inert template element and move its children into
       // `element` but such that `element`'s own children are not replaced
-      const tmpl = element.ownerDocument.createElement('template');
+      const tmpl =
+        document.createElementNS('http://www.w3.org/1999/xhtml', 'template');
```
zbraniecki: that means that I can use HTML Template element for overlays in XUL.

src/bindings/overlay.js
```diff
@@ -155,7 +157,7 @@ function getNthElementOfType(context, element, index) {
   let nthOfType = 0;
   for (let i = 0, child; child = context.children[i]; i++) {
     if (child.nodeType === child.ELEMENT_NODE &&
-        child.tagName === element.tagName) {
+        child.tagName.toLowerCase() === element.tagName.toLowerCase()) {
```
zbraniecki: XHTML and XUL report tag name upper case, HTML5 lower case.

src/intl/context.js
```diff
@@ -1,16 +1,22 @@
...
+
+Cu.import("resource://gre/modules/XPCOMUtils.jsm");
+XPCOMUtils.defineLazyModuleGetter(this, "L20nParser",
+                                  "resource://gre/modules/L20nParser.jsm");
+
 export class MessageContext {
-  constructor(lang, { functions = {} } = {}) {
-    this.lang = lang;
+  constructor(locales, { functions = {}, args = {} } = {}) {
+    this.locale = Array.isArray(locales) ? locales[0] : locales;;
```
zbraniecki: If we want to go for Intl API, it's locales and this.resolvedOptions().locale, not `lang`.

src/intl/context.js
```diff
@@ -1,16 +1,22 @@
...
     this.functions = functions;
+    this.args = args;
```
zbraniecki: I'd like to keep the context data for now since I see use cases for it and it costs nothing.

src/intl/resolver.js
```diff
@@ -257,6 +257,9 @@ function* toString(entity) {
 }
 
 export function format(ctx, args, entity) {
+  if (entity === undefined) {
+    return ['', []];
+  }
```
zbraniecki: I had a problem where a missing entity would break something. Need to test what was breaking to understand if that's the right fix.

src/intl/types.js
```diff
@@ -67,7 +76,7 @@ export class FTLKeyword extends FTLBase {
 export class FTLList extends Array {
   toString(ctx) {
     const lf = ctx._memoizeIntlObject(
-      L20nIntl.ListFormat // XXX add this.opts
+      ListFormat // XXX add this.opts
```
zbraniecki: what stops us from putting `other.opts` here?

src/lib/format.js
```diff
@@ -14,7 +14,7 @@ export function keysFromContext(ctx, keys, method) {
 export function valueFromContext(ctx, id, args) {
   const entity = ctx.messages.get(id);
 
-  if (!entity) {
+  if (entity === undefined) {
```
zbraniecki: entity may be defined as an empty string (`key = ""`)

src/lib/format.js
```diff
@@ -24,14 +24,18 @@ export function valueFromContext(ctx, id, args) {
...
       [new L10nError(`Unknown entity: ${id}`)]
     ];
   }
 
-  const [value] = ctx.format(entity, args);
+  let value = null;
+
+  if (typeof entity === 'string' || Array.isArray(entity) || entity.val !== undefined) {
+    value = ctx.format(entity, args)[0];
+  }
```
zbraniecki: entity may have an empty value:
```
key =
  [html/title] Foo
```

src/bindings/dom.js
stasm: What are the possible failures?

src/bindings/dom.js
stasm: :+1:

src/bindings/overlay.js
stasm: Yeah, I'm not opposed but also don't want to optimize dir layout too early.  I'm okay with having this `button` here even for HTML for now.

src/bindings/overlay.js
stasm: IIRC, we used `element.ownerDocument` to make sure principals where set correctly and methods like `importNode` could work.  I also think this was related to Gaia, although I don't quite remember the details.

src/bindings/overlay.js
stasm: <3

src/intl/context.js
stasm: `lang` comes from the `ResourceBundle` and in my mind should always be a single language code.  Intl's locales can be used in the  language negotiation process which creates a list of Resource Bundes.

src/bindings/overlay.js
zbraniecki: the problem is that XUL doesn't have ownerDocument and we don't want to create XUL element in XUL. We want to create HTML element in XUL :)

zbraniecki: (because only HTML has Template Tag)

src/intl/context.js
stasm: A `MessageContext` can only ever be in a single language because it holds translations for just one language.

zbraniecki: I doubt we'll be able to push MessageContext for standardization with other signature than "locales, options". And if we don't want to push for standardization then maybe we should not put it in Intl?

I do believe that we'll need some (public) API to negotiate locales for multiple formatters - like give me the best locale fallback chain for Messages/Units/DateTime. And we should aim to polyfill that.

src/intl/context.js
stasm: I'd like to set the right expectations.  If a dev set ctx.args = { name: 'Mary' } and than the name changes, is it understood that they also need to manually reformat all messages?

If we don't need this immediately, I'd leave it out.  It's easy to add in this form in the future, but we might end up doing something more for it, too.

src/intl/context.js
zbraniecki: Exactly as DateTimeFormat is in a single language, and UnitFormat and NumberFormat. But you pass locales to it's constructor.

src/intl/resolver.js
stasm: I've seen this too.  I'll investigate and maybe fix on master.  This looks like a right fix.

src/intl/types.js
stasm: It was easier to implement `FTLList` as a subclass of `Array` for now.  This meant that we couldn't use the second argument to the constructor for `ops` because it's the second element of the array.  (It's dangerous to override the `Array`'s constructor because it is still used in other places internally, e.g. when calling `.map()`).  I think the easiest will be to add a static `FTLList.create(iterable, opts)` in plce of the currently-usd `FTLList.from` (inherited from `Array`; also can't override the second arg).  I just didn't get around to doing it.

Or we could not sublass `Array` and implement custom iteration and `.push()` and `.map()`.  or maybe re-use `Array.prototype.push` and `Array.prototype.map`;  I'd need to try.

Do we need `opts` right now or can we add them later on?

src/lib/format.js
stasm: Good catch, I typed too fast :)

src/lib/format.js
stasm: Should this be fixed in the resolver rather than here?  I can give it a closer look tomorrow on master.

src/intl/context.js
stasm: It's different. `DateTimeFormat` takes a list of locales because it negotiates the best language given the installed ICU dataset.  `MessageContext` doesn't take its translations from the ICU;  they're passed to it via `.addMessage()`.  It isn't possible for `MessageContext` to end up being in a different language than the one supplied.

And yes, we want to standardize it.

src/lib/format.js
stasm: What if there's no value but one of the traits has the asterisk?

    key =
     *[html/title] Foo

I think it would be safest to just run this through the resolver and fix the resolver to know about `null` values?

zbraniecki: sure, I just fixed it because it broke my XUL file :) I'm happy to get a better fix in master

src/intl/resolver.js
stasm: Actually, `entityFromContext` should prevent errors here, so I wonder why you saw them at all. Did you try to call call `ctx.format` directly somewhere?

src/lib/format.js
stasm: I fixed this in https://github.com/l20n/l20n.js/commit/ed19609d1cda0fd4013c2a057c565bd3623e23fe by adding `MessageContext.formatToPrimitive` which can return string or null values.

src/intl/resolver.js
stasm: I added the guard in https://github.com/l20n/l20n.js/commit/0cbc5694bb6654f7bc135bf590c19b44364e6f44

src/lib/format.js
stasm: Fixed in https://github.com/l20n/l20n.js/commit/c794525f3c1e453d7f8c295514f04fd78435a862

PR #126, opened 2016-05-24
Gecko dev

zbraniecki:
This is ready for review @stasm

### DIFF
--- a/src/ftl/ast/parser.js
+++ b/src/ftl/ast/parser.js
@@ -143,7 +143,7 @@ class ParseContext {
       members = this.getMembers();
     } else if (value === null) {
       throw this.error(
-  `Expected a value (like: " = value") or a trait (like: "[key] value")`);
+  'Expected a value (like: " = value") or a trait (like: "[key] value")');
     }
 
     return new AST.Entity(id, value, members, comment);
--- a/src/ftl/ast/parser.js
+++ b/src/ftl/ast/parser.js
@@ -143,7 +143,8 @@ class ParseContext {
       members = this.getMembers();
     } else if (value === null) {
       throw this.error(
-  'Expected a value (like: " = value") or a trait (like: "[key] value")');
+        'Expected a value (like: " = value") or a trait (like: "[key] value")'
+      );
     }
 
     return new AST.Entity(id, value, members, comment);
--- a/src/runtime/gecko/IntlMessageContext.jsm
+++ b/src/runtime/gecko/IntlMessageContext.jsm
@@ -2,4 +2,13 @@ import { MessageContext } from '../../intl/context';
 
 this.EXPORTED_SYMBOLS = ['MessageContext'];
 
+Components.utils.import('resource://gre/modules/IntlListFormat.jsm');
+Components.utils.import('resource://gre/modules/IntlPluralRules.jsm');
+Components.utils.import('resource://gre/modules/IntlRelativeTimeFormat.jsm');
+
+Intl.PluralRules = PluralRules;
+Intl.ListFormat = ListFormat;
+Intl.RelativeTimeFormat = RelativeTimeFormat;
+
+
 this.MessageContext = MessageContext;
--- a/src/runtime/gecko/IntlMessageContext.jsm
+++ b/src/runtime/gecko/IntlMessageContext.jsm
@@ -1,0 +1,34 @@
+import { MessageContext } from '../../intl/context';
+
+
+this.EXPORTED_SYMBOLS = ["MessageContext", "MessageArgument"];
+
+
+function MessageArgument(formatter, options, values) {
+  if (values === undefined) {
+    return (values) => {
+      if (values === undefined) {
+        throw new Error('Cannot resolve Argument without a value');
+      }
+      if (!Array.isArray(values)) {
+        values = [values];
+      }
+      return {
+        formatter,
+        options,
+        values
+      };
+    };
+  }
+  if (!Array.isArray(values)) {
+    values = [values];
+  }
+  return {
+    formatter,
+    options,
+    values
+  };
+}
+
+this.MessageContext = MessageContext;
+this.MessageArgument = MessageArgument;
--- a/src/runtime/gecko/IntlMessageContext.jsm
+++ b/src/runtime/gecko/IntlMessageContext.jsm
@@ -1,34 +1,5 @@
 import { MessageContext } from '../../intl/context';
 
-
-this.EXPORTED_SYMBOLS = ["MessageContext", "MessageArgument"];
-
-
-function MessageArgument(formatter, options, values) {
-  if (values === undefined) {
-    return (values) => {
-      if (values === undefined) {
-        throw new Error('Cannot resolve Argument without a value');
-      }
-      if (!Array.isArray(values)) {
-        values = [values];
-      }
-      return {
-        formatter,
-        options,
-        values
-      };
-    };
-  }
-  if (!Array.isArray(values)) {
-    values = [values];
-  }
-  return {
-    formatter,
-    options,
-    values
-  };
-}
+this.EXPORTED_SYMBOLS = ['MessageContext'];
 
 this.MessageContext = MessageContext;
-this.MessageArgument = MessageArgument;
--- a/src/runtime/gecko/L20n.jsm
+++ b/src/runtime/gecko/L20n.jsm
@@ -1,5 +1,0 @@
-import { ResourceBundle } from './resourcebundle';
-
-this.EXPORTED_SYMBOLS = ["ResourceBundle"];
-
-this.ResourceBundle = ResourceBundle;
--- a/src/runtime/gecko/L20n.jsm
+++ b/src/runtime/gecko/L20n.jsm
@@ -1,0 +1,5 @@
+import { ResourceBundle } from './resourcebundle';
+
+this.EXPORTED_SYMBOLS = ["ResourceBundle"];
+
+this.ResourceBundle = ResourceBundle;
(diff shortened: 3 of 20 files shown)


tools/check-strings.js
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env node
```
stasm: remove this file?

tools/extract-strings.js
```diff
@@ -23,12 +23,14 @@ function trimString(str) {
 function extractFromHTML(err, data) {
   const res = new AST.Resource();
 
-  const $ = cheerio.load(data.toString());
+  const $ = cheerio.load(data.toString(), {
```
stasm: remove this file as well?

src/runtime/gecko/io.js
```diff
@@ -0,0 +1,27 @@
+import { L10nError } from '../../lib/errors';
```
stasm: `L10nError` isn't used in this file.  Please lint the whole PR.

src/runtime/gecko/l20n-html.js
```diff
@@ -0,0 +1,56 @@
+import { prioritizeLocales } from '../../intl/locale';
+import { contexts, Localization } from '../../bindings/html';
+import { documentReady, getResourceLinks, getMeta } from './util';
+import { keysFromContext, valueFromContext } from '../../lib/format';
+
+Components.utils.import('resource://gre/modules/Services.jsm');
+Components.utils.import('resource://gre/modules/L20n.jsm');
+Components.utils.import('resource://gre/modules/IntlMessageContext.jsm');
+
+const functions = {
```
stasm: perhaps move this into `src/runtime/gecko/functions`?

src/runtime/gecko/l20n-html.js
```diff
@@ -0,0 +1,56 @@
...
+      lang => new ResourceBundle(lang, resIds)
+    );
+  });
+}
+
+function createContext(lang) {
+  return new MessageContext(lang, { functions });
+}
+
+document.l10n = new Localization(document, requestBundles, createContext);
```
stasm: I think I preferred an approach in which `MessageContext` is exposed on a global namespace, like `Intl`, rather than having to pass `createContext` around.  What where the reasons for not doing it?

src/runtime/gecko/util.js
```diff
@@ -0,0 +1,41 @@
+export { documentReady, getResourceLinks, getMeta } from '../web/util';
+import { getLangRevisionMap, getLangRevisionTuple } from '../web/util';
+
+export function getXULResourceLinks(doc) {
+  return Array.prototype.map.call(
+    doc.querySelectorAll('messagebundle'),
+    el => el.getAttribute('src'));
+}
+
+export function getXULMeta(doc) {
```
stasm: Do we still need to inspect `<meta>` elements in Gecko runtime?

src/runtime/gecko/L20nParser.jsm
```diff
@@ -0,0 +1,5 @@
+import { default as L20nParser } from '../../ftl/entries/parser';
+
+this.EXPORTED_SYMBOLS = ["L20nParser"];
+
+this.L20nParser = L20nParser;
```
stasm: Looking at the `makefile`, this file isn't used.

src/runtime/gecko/L20n.jsm
```diff
@@ -0,0 +1,5 @@
+import { ResourceBundle } from './resourcebundle';
+
+this.EXPORTED_SYMBOLS = ["ResourceBundle"];
+
+this.ResourceBundle = ResourceBundle;
```
stasm: I don't think there's a good enough reason to put resource bundles in their own file.  Could simply `import` them in each of the entry points so that rollup bundles them in the dist file?

src/runtime/gecko/IntlMessageContext.jsm
```diff
@@ -0,0 +1,14 @@
+import { MessageContext } from '../../intl/context';
+
+this.EXPORTED_SYMBOLS = ['MessageContext'];
+
+Components.utils.import('resource://gre/modules/IntlListFormat.jsm');
+Components.utils.import('resource://gre/modules/IntlPluralRules.jsm');
+Components.utils.import('resource://gre/modules/IntlRelativeTimeFormat.jsm');
```
stasm: Are these files part of our build?

src/runtime/gecko/IntlMessageContext.jsm
```diff
@@ -0,0 +1,14 @@
...
+
+Intl.PluralRules = PluralRules;
+Intl.ListFormat = ListFormat;
+Intl.RelativeTimeFormat = RelativeTimeFormat;
```
stasm: I remember you wanted to use lazy getters for these.  I'm okay with not using them for now; just interested to hear what troubles you had with them.

src/runtime/gecko/util.js
```diff
@@ -0,0 +1,41 @@
+export { documentReady, getResourceLinks, getMeta } from '../web/util';
+import { getLangRevisionMap, getLangRevisionTuple } from '../web/util';
+
+export function getXULResourceLinks(doc) {
+  return Array.prototype.map.call(
+    doc.querySelectorAll('messagebundle'),
```
stasm: Have you considered using `<html:link>` elements instead of `<messagebundle>`?  Right now it looks like `gteResourceLinks` is the only difference between `l20n-xul` and `l20n-html` and I'm wondering if we could actually just use one.

stasm:
Great work, @zbraniecki.  I'm happy to see this coming along nicely.  The biggest open question to me is how we expose `MessageContext` to `Localization` objects.  I would prefer to no pass a `createContext` function around if possible.

src/runtime/gecko/l20n-html.js
stasm: Ah, but it looks like we'd need to pass `functions` anyways.  I see.

src/runtime/gecko/l20n-html.js
stasm: if we manage to merge `l20n-xul` and `l20n-html` into a single file then it's OK for this to be here

src/runtime/gecko/l20n-html.js
stasm: Can we still try to expose `MessageContext` on `Intl`?

stasm:
If I understand correctly, if we don't pass `createContext`, we still need to pass the runtime-specific `functions` into `Localization`.  So that's fine.

I was trying to think about other ways of exposing `MessageContext` but in the end I'm okay with your approach.  Thanks!

src/runtime/gecko/l20n-html.js
zbraniecki: we can, but it's not used anywhere :)

src/runtime/gecko/util.js
zbraniecki: for now yes, we don't have any other place to get info about languages available for the given files yet.

src/runtime/gecko/util.js
zbraniecki: I'm not sure if it'll fly with people to use add another NS to every XUL file for the sake of our internal codebase size

tools/check-strings.js
zbraniecki: why? It's an important addition that allows me to check if strings are used.

tools/extract-strings.js
zbraniecki: why?

src/runtime/gecko/IntlMessageContext.jsm
zbraniecki: no. they come from mozIntl.

src/runtime/gecko/IntlMessageContext.jsm
zbraniecki: yes, I'll be switching to lazy getters once I start working on perf.

src/runtime/gecko/L20n.jsm
zbraniecki: I was trying to make space for the future Env singleton for caching purposes. Do you think you will not want to go there?

zbraniecki:
Updated to your comments. Lmk

src/ftl/ast/parser.js
```diff
@@ -143,7 +143,7 @@ class ParseContext {
       members = this.getMembers();
     } else if (value === null) {
       throw this.error(
-  `Expected a value (like: " = value") or a trait (like: "[key] value")`);
+  'Expected a value (like: " = value") or a trait (like: "[key] value")');
```
stasm: nit: indentation; the closing `)` should go into the next line

src/runtime/gecko/io.js
```diff
@@ -0,0 +1,24 @@
...
+    const req = Components.classes['@mozilla.org/xmlextras/xmlhttprequest;1']
+                      .createInstance(Ci.nsIXMLHttpRequest)
+
+    req.mozBackgroundRequest = true;
+    req.overrideMimeType('text/plain');
+    req.open('GET', url, true);
+
+    req.addEventListener('load', () => {
+      if (req.status === HTTP_STATUS_CODE_OK) {
+        resolve(req.responseText);
```
stasm: we should also reject the promise if the request fails

src/runtime/gecko/l20n-xul.js
```diff
@@ -0,0 +1,56 @@
...
+      default:
+        return 'other';
+    }
+  }
+};
+
+function requestBundles(requestedLangs = navigator.languages) {
+  return documentReady().then(() => {
+    const defaultLang = 'en-US';
+    const availableLangs = ['en-US'];
```
stasm: No `pl` here?

zbraniecki: I didn't provide 'pl' translation for XUL file yet.

src/runtime/gecko/io.js
zbraniecki: done

stasm:
r+, thanks @zbraniecki!

PR #127, opened 2016-05-25
GeckoLocalization

stasm:
Changes required to make the demo for MozLondon work.

### DIFF
--- a/src/runtime/gecko/localization.js
+++ b/src/runtime/gecko/localization.js
@@ -1,0 +1,57 @@
+import { Localization, contexts, translateDocument }
+  from '../../bindings/html';
+import { valueFromContext } from '../../lib/format';
+
+// Services.jsm is imported in entry points: l20n-html and l20n-xul
+
+const functions = {
+  OS: function() {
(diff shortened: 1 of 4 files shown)


stasm:
@zbraniecki Do you think we could merge this now?

src/runtime/gecko/localization.js
```diff
@@ -0,0 +1,57 @@
+import { Localization, contexts, translateDocument }
+  from '../../bindings/html';
+import { valueFromContext } from '../../lib/format';
+
+// Services.jsm is imported in entry points: l20n-html and l20n-xul
```
zbraniecki: Why not load it here? rollup gets confused about undefined variable `Services` and can reassign it's name...

zbraniecki:
r+

src/runtime/gecko/localization.js
stasm: We need `Services` for `Service.obs.addObserver` in the entry points.  Can I just import it twice?  It's memoized anyways, right?

PR #128, opened 2016-06-08
Separate Localization and LocalizationObserver (for XBL)

stasm:
This PR introduces the `LocalizationObserver` class which manages DOM mutations and listens to the `languagechange` events.  It's available as `document.l10n` and is a subclass of `Map`.  Individual `Localization` objects are available via `document.l10n.get(name)` where `name` should be a string or a symbol.  The default unnamed `Localization` can be identified by `Symbol.for('anonymous l10n')`:

    document.l10n.get(Symbol.for('anonymous l10n')).formatValues(…);

…although I'd encourage people to give their `Localizations` names instead:

    document.l10n.get('main').formatValues(…);

The main idea behind the changes is that `Localization` objects manage resources and fallback, and `LocalizationObserver` manages mutations, translating the DOM (also anonymous in XBL's case) and setting `lang` and `dir` on localizable roots.

Each root (a DOM element) has one corresponding `Localization` object which will set the attributes according to the state of the language negotiation in the object.  Children of the root are free to use other `Localization` objects to provide translations for them, however.  This can be achieved via the `data-l10n-bundle` attribute.

A few things aren't implemented yet, but this is generally ready for review right now.

  - `translate*` functions won't currently look for anonymous content in chrome-privileged code,
  - `name` attributes on `links` aren't supported; the default unnamed `Localization` is always assumed.


stasm:
@zbraniecki Happy to talk you through the changes when you're available.

stasm:
I'd like to point out an important consequence of the named bundles approach. Because the names are defined in the DOM, we can't create `Localization` objects right away—we need to wait for `DOMContentLoaded` first.  Initially we don't even know just _how many_ objects we'll want to create.

I'm fine with this limitation because the upside is a cleaner API which allows to group and structure the resources better. We could consider adding a `document.l10n.format*` wrappers which internally wait for the default unnamed bundle to be created, but in fact I'd like to not to this for now and experiment with porting Gecko code without it.  We all want to use `data-l10n-*` attributes instead after all :)

stasm:
I changed `Symbol.for('anonymous l10n')` to `main` because I'd like to evoke the meaning of the _default_ localization rather than an unnamed one.  If the `name` attribute isn't defined, `name="main"` is implied.

I also implemented defining `name` on `links` in the HTML API.

stasm:
`ChromeLocalizationObserver` now uses `document.getAnonymousNodes` to translate anonymous nodes attached to the bound root.

stasm:
I refactored `getElementsTranslation` to take advantage of `formatEntities` which takes an array of keys.  This fixes the WIP version which I introduced earlier on the branch in https://github.com/l20n/l20n.js/pull/128/commits/e73028b57351326795a6bbc45ca1d2509dfac449.

src/bindings/dom/html.js
```diff
@@ -0,0 +1,65 @@
...
+};
+
+export class HTMLLocalization extends Localization {
+  overlayElement(element, translation) {
+    return overlayElement(this, element, translation);
+  }
+
+  // XXX the allowed list should be amendable; https://bugzil.la/922573
+  isElementAllowed(element) {
+    return allowed.elements.indexOf(element.tagName.toLowerCase()) !== -1;
```
Pike: we should only lowercase in HTML documents for HTML Elements.

Also, tagName or localName? Do we have prefixed markup in l10n?

stasm: We should allow for localization of XUL elements defined in XBL, e.g. `<xul:label/>`.  `localName` sounds like a way to go, perhaps together with `namespaceURI`.  I'd make these changes only in `XULLocalization` for now.

zbraniecki:
r+, great job stas!

PR #130, opened 2016-06-24
Update Polish decimal case

irisgau:
Covers polish plural case with decimals. For example, 0,3⁩ sekundy and 4,3⁩ sekundy.


stasm:
Thanks, @irisgau!

PR #138, opened 2016-09-06
Add FTLASTSerializer to node js module

ta2-1:
Let me to describe why I need this change. 
I'd like to use a ready l20n.js node js module but I can't do it now.

I convert arrays like:
```
["{ $num } new notification", "{ $num } new notifications"]
```
into l20n value
```
{ PLURAL($num) -> 
  [one] { $num } new notification.
 *[other] { $num } new notifications.
}
```

I need `FTLASTSerializer.dumpPattern()` method which is not in node.js bundled module.


stasm:
This makes sense, thanks for the PR!

PR #140, opened 2016-09-09
Disallow whitespace at the end of keyword

zbraniecki:
Need to update tests.


zbraniecki:
Pushed as https://github.com/l20n/l20n.js/commit/a42bbface9efe5f7ce80abde1648944a479b4a87

PR #142, opened 2016-09-15
Bug 1303148 - Use a documentation generator

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1303148


zbraniecki:
One thing I lack here is explanation of arguments, but I see you used `@returns` in a couple places so I assume we can start from that and add some semantically readable description of arguments that those functions take?

stasm:
Yes, absolutely.  For now I simply copied what we had in `docs` already.  All of `src/lib/dom` is on my list so I'll make edits where needed.

stasm:
Got an r+ from @zbraniecki on irc.

PR #143, opened 2016-09-17
Using L20n To Localize XUL Documents

stasm:
@zbraniecki what do you think?


zbraniecki:
yeah, looks good. I'll take a stab at some edits later, but you should land it :)

stasm:
Thanks!

PR #146, opened 2016-09-22
Use a high workload FTL file for make perf by default

stasm:
The current `example.ftl` file used with `make perf` is an exact copy of the strings we ported in [Bug 1288406 - Port Firefox's menu bar to FTL](https://bugzilla.mozilla.org/show_bug.cgi?id=1288406).  While it's representative of the majority of strings in Firefox, it makes it hard to test the true performance of the resolver -- almost all translations are simple strings.

I'd like to land a new file, `workload-high.ftl`, and use it by default with `make perf`.  In cases where we'd want to test the performance of the real manubar strings we can edit the benchmark files by hand to make the load `workload-low.ftl` instead.

Here's the difference between `master` and this PR:

```
parseFTL: 
  mean:   10889.63 (+24%)
  stdev:  1267.51
  sample: 30
parseFTLEntries: 
  mean:   4197.97 (+38%)
  stdev:  450.28
  sample: 30
format: 
  mean:   19144.97 (+1562%)
  stdev:  1960.97
  sample: 30
```

@zbraniecki r?


zbraniecki:
could you actually make it possible to call make perf with an argument to test low vs high?

stasm:
I probably could introduce yet another env var that `make perf` recognizes, `PERF_WORKLOAD=high` or something.  I'm not sure how to pass its value from `tools/perf/test.js` into the benchmark files. Probably `process.env` for node, [options](https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Introduction_to_the_JavaScript_shell#options.28.5Boption_....5D.29) for jsshell and I don't know what for d8 (I'd remove it tbh).  Do you think this is worth the effort?

zbraniecki:
well, I heavily use this tool right now to try to optimize the parser for larch perf so I benefit from the correlation of what we parse there and in make perf.

So my answer is yes, I think it's worth it and if you won't write it before landing, i'll have to :)

stasm:
Can I remove `benchmark.d8.js`?

zbraniecki:
why do you want to do that? It gives us ability to test against v8.

stasm:
it's one more file to maintain and it adds more work for us, for instance in this case where I need to pass envvars into the script. We also have `benchmark.node.js` for testing a relatively modern version of v8. And it doesn't help us with the goal of landing l20n.js on larch at all.

stasm:
I also just remembered that the measurements done by `benchmark.d8.js` were never very accurate because it used a regular `Date.now()` which has a resolution of 1 ms, while both `benchmark.jsshell.js` and `benchmark.node.js` use high-resolution timers (jsshell's `dateNow` reports milliseconds with decimals and node's `process.hrtime` reports tuples of milliseconds and nanoseconds).

Let's use `benchmark.node.js` to keep performance on v8 in check.  Node v6.5.0 uses V8 5.1.281.75 which shipped in Chrome 51 (the current version of Chrome is 53).

zbraniecki:
> Let's use benchmark.node.js to keep performance on v8 in check

Node and d8 have different performance characteristics and we can only run many instrumentations against d8 (like, jit profiling, deopts profiling, inline profiling etc.)

So if I was forced to keep one, I'd choose d8 over node, but I'm still not sure why are you trying to limit it.
Is it because you don't use it? I do...

stasm:
I'm afraid you're using something that doesn't report very accurate results.

I'm not able to get jsshell to read env vars.  Despite the documentation, the `environment` global doesn't work. I'll simply land a new file (`workload-high`) and will edit the benchmarks by hand when I need it.  I'll have a new version of this PR ready in no time.

stasm:
@zbraniecki r?

zbraniecki:
r+, thanks! I still hope we can get an env param for that eventually (and increase the resolution in d8, I'll look into that)

stasm:
Thanks!

For reference, I was trying to use https://developer.mozilla.org/en-US/docs/Mozilla/Projects/SpiderMonkey/Shell_global_objects#Global_objects in `benchmark.jsshell.js` but `environment` just gave me a `ReferenceError`.

PR #150, opened 2016-09-29
Bug 1306262 - Wrapping in Bidi Isolation marks should happen in bindings

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1306262

### DIFF
--- a/src/intl/resolver.js
+++ b/src/intl/resolver.js
@@ -35,10 +35,6 @@ import { FTLType, FTLNone, FTLNumber, FTLDateTime, FTLKeyword, FTLList }
   from './types';
 import builtins from './builtins';
 
-// Unicode bidi isolation characters.
-const FSI = '\u2068';
-const PDI = '\u2069';
-
 // Prevent expansion of too long placeables.
 const MAX_PLACEABLE_LENGTH = 2500;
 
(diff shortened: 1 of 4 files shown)


src/intl/resolver.js
```diff
@@ -329,9 +325,9 @@ function Pattern(env, ptn) {
             `(${str.length}, max allowed is ${MAX_PLACEABLE_LENGTH})`
           )
         );
-        result += FSI + str.substr(0, MAX_PLACEABLE_LENGTH) + PDI;
+        result += str.substr(0, MAX_PLACEABLE_LENGTH);
```
zbraniecki: since this is a hot path, can you test array and `.join` at the end instead of concatenation?

src/lib/dom/base.js
```diff
@@ -299,3 +299,76 @@ function equal(bundles1, bundles2) {
...
+ * }} prev
+ * @returns {{
+ *   errors: Array<Error>,
+ *   hasErrors: boolean,
+ *   translations: Array<{value: string, attrs: Object}>
+ * }}
+ * @private
+ */
+function entitiesFromContext(ctx, keys, prev) {
+  return keysFromContext(entityFromContext, sanitizeArgs, ctx, keys, prev);
```
zbraniecki: not sure how perf-critical it is, but did you consider partial function here instead of bound?

src/intl/resolver.js
stasm: There's no difference between `+=` and `.join()`.  Also no difference between using `FSI` and `PDI` here and not.  I tested using `workload-high.ftl`.

src/lib/dom/base.js
stasm: Can you explain what you mean? Isn't what I did here a partially applied function?

zbraniecki:
```javascript
const entitiesFromContext = keysFromContext.bind(null, entityFromContext, sanitizeArgs);
```

Does it make sense?

stasm: This is functionally exactly the same, right?  I think I prefer to use the `function` declaration and keyword.

PR #153, opened 2016-10-12
Second pass on hacking.rst

stasm:
I applied the feedback we got on the first iteration of the Hacking documentation.

@Pike, @zbraniecki — r?

### DIFF
--- a/docs/hacking.rst
+++ b/docs/hacking.rst
@@ -2,273 +2,263 @@
 Hacking
 =======
 
-L20n library is a combination of three layers of APIs built on top of each other.
-The intention is for the user to be able to use just the first or first and the second
-layer separately.
+This document is intended as a guide for people interested in contributing to 
+l20n.js or curious about its inner workings.  It presents an overview of the 
+architecture of l20n.js.
 
-Each layer has different external dependencies and is intended to tie better into the
-host and client environments.
+The l20n.js library is built as a combination of three layers of APIs stacked on 
+top of each other.  This modularization allows l20n.js to support many 
+different environments and platforms.
 
-Intl API
-========
+Intl
+    The low-level API providing single-purpose building blocks for l20n.js.
 
-The bottom layer is completely dependency-free and environment independent.
-It's shaped similarly to `ICU MessageFormat`_ API.
+Localization
+    The main l20n.js class responsible for language negotiation, resource 
+    loading and the fallback strategy.
+
+LocalizationObserver
+    The manager class used to bind to the host environment.
+
+Each layer has different external dependencies and is intended to tie better 
+into the host and client environments.  In addition to the above three layers 
+there's also the so-called runtime code which is responsible for instantiating 
+the proper classes, attaching event listeners and triggering the initial 
+translation of the UI.
+
+
+The Intl API
+============
+
+L20n.js builds on top of the JavaScript Intl API (ECMA 402).  L20n.js uses 
+``Intl.NumberFormat`` to format numbers in a locale-aware manner and 
+``Intl.DateTimeFormat`` to format dates.  Other features are implemented as 
+extensions to the Intl API modeled after the existing ECMA specifications.  We
+work with ECMA to enhance the APIs which address the needs of L20n and other 
+libraries needing access to internationalized data.
+
+``ListFormat``
+    Provides a way to format lists and enumerations in a locale-aware manner.  
+    Proposed to ECMA 402 as `Intl.ListFormat`_.
+
+``PluralRules``
+    Returns CLDR plural categories (zero, one, two, few, many, other) using the
+    given locale's plural rule. Proposed to ECMA 402 as `Intl.PluralRules`_.
+
+``MessageContext``
+    Formats translations to strings allowing for interpolating external 
+    arguments passed in by the developer and some more logic if needed.
+
 
 Entity
 ------
 
-A single localization unit in L20n vocabulary is called an entity.
-An entity is an ``Object`` with three fields:
+A single localization unit in ``MessageContext`` is called an entity.  An entity 
+has an identifier, a value and traits.
 
-:id:
-    An identifier
+id
+    The unique identifier of the entity.
 
-:value:
-    A value that may be a simple or more complex computable string.
+value
+    The entity's value. It may be null.
 
-:traits:
+traits
     A list of key-value pairs that can be providing additional data for the entity like
-    attributes or variants.
+    attributes or variants. It may be empty.
 
-Entities are encoded in a source format named FTL. L20n provides two parsers for the
-format. The full AST parser is available for tools and external users who need
-access to the full AST of the file.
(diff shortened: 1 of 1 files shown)


docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
...
-The intention is for the user to be able to use just the first or first and the second
-layer separately.
+The L20n library is built as a combination of three layers of APIs stacked on 
+top of each other.  This modularization allows l20n.js to support many 
+different environments and platforms.
 
-Each layer has different external dependencies and is intended to tie better into the
-host and client environments.
+:ECMA 402 Intl
+    The low-level API providing L20n's primitives and building-blocks.
```
zbraniecki: So, I'm concerned about using the term "ECMA" here as misleading. Some of those building blocks are not yet in ECMA and I'd prefer not to pretend that this is ECMA.
In particular the most important class - MessageContext - has not been even suggested to ECMA yet.

docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
...
+translation of the UI.
+
+
+The Intl API
+============
+
+At the heart of l20n.js is the low-level ECMA 402's Intl API.  L20n.js uses 
+``Intl.NumberFormat`` to format numbers in a locale-aware manner and 
+``Intl.DateTimeFormat`` to format dates.  Other features are provided by our custom
+extensions to the Intl API, modeled after the existing ECMA specifications:
```
zbraniecki: Instead, I'd suggest to put here a note that states that we intend to work with ECMA on making all APIs on this layer be aligned with ECMA402.

docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
...
-The bottom layer is completely dependency-free and environment independent.
-It's shaped similarly to `ICU MessageFormat`_ API.
 
 Entity
 ------
 
-A single localization unit in L20n vocabulary is called an entity.
-An entity is an ``Object`` with three fields:
+A single localization unit in ``MessageContext`` is called an entity.  An entity 
+has an identifier, a value and traits.
```
zbraniecki: Not sure if it's something to fix but you mention MessageContext before you define it.

docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
...
-            an object with environmental function that are going to be available
-            from within the entities.
-    
-:addMessages(source):
-    takes a localization file source (at the moment FTL file format is supported),
-    parses the source into localization messages and stores them.
+The translations are considered untrusted. ``MessageContext`` has been 
+designed to be resilient to formatting errors.  If a translation contains 
+errors, ``MessageContext`` will try to salvage as much of the translation as 
+possible and will still return a string.
```
zbraniecki: This reads a bit weird. The `The translations are considered untrusted` being the first sentence of the paragraph feels odd.
Maybe put it further down in the paragraph or explain a bit: `L20n takes an approach of assuming that localization resources do not go through a security review process and may be coming from a third-party, thus the translation have to be considered untrusted`.

docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
...
 
-:format(entity, args, errors):
-    fromats a message and returns it as an Object.
+For performance reasons formatting errors aren't returned together with the 
+formatted string.  The ``format`` method of ``MessageContext`` takes an 
+additional parameter, ``errors``, which if provided must be an Array to which 
+errors will be pushed.
```
zbraniecki: The first sentence sounds like returning error with the formatted string is the default programming pattern and we need to justify not following it - feels defensive :)
I believe you can do without this sentence all together.

docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
...
-``ResourceBundle`` API:
+In Web-like runtimes the ``Localization`` API is fully asynchronous.  This 
+allows for loading fallback resources lazily in case of errors in the first 
+language.
 
-:lang:
-    Language tag associated with the bundle
+The ``Localization`` class can be thought of as a multi-lingual, asynchronous 
+wrapper on top of multiple ``MessageContext`` instances which handles lazy-IO 
+and the fallback strategy.
```
zbraniecki: `language fallback` - I believe that it's worth pointing out what is fallbacking

docs/hacking.rst
Pike: Similarily, I think we should call this something like "L20n builds on top of JavaScript Intl APIs. We also work with ECMA to enhance these APIs where there's overlap between the needs of L20n and other libraries needing access to internationalized data."

docs/hacking.rst
stasm: I think that's okay because ``MessageContext`` is already mentioned in the list above.

docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
...
+translations to the bundle's language.
 
-:resIds:
-    List of resources for the bundle.
+The ``Localization`` constructor takes two runtime-specific functions as 
+arguments, ``requestBundles`` and ``createContext``. This allows the user to 
+provide different ways to build bundles, use different IO and language 
+negotiation methods, as well as pass custom functions to the ``MessageContext``
+constructor.  The ``Localization`` class itself doesn't know anything about the
+environment it runs in and doesn't provide any special methods to localize it.
```
Pike: "environment" and "localize it" are a tad underdefined. Maybe this is easier if you start with the sentence below right away, somewhat like:

The ``LocalizationObserver`` adapts the Localization to APIs that are properly suited for programming environment.

or so?

docs/hacking.rst
```diff
@@ -2,273 +2,256 @@
 Hacking
 =======
```
Pike: Maybe add a quick paragraph on top to clarify the target audience for this doc?

docs/hacking.rst
stasm: I think I'll just mention that `Localization` is env-agnostic.  The first sentence in the next section explains this better anyways.

docs/hacking.rst
```diff
@@ -72,21 +79,23 @@ MessageContext
 ``MessageContext`` is a synchronous single-locale map of translations.  It can 
 format translations to strings.
 
-The translations can be added with the ``addMessages`` method.  They are 
-internally parsed into a lazy intermittent form and then formatted only on 
-retrieval.  The ``format`` method allows the user to format them using 
-environmental variables, user-provided arguments, other messages in the current 
-context and any other syntax that the format allows.
+Translations must be defined in the FTL syntax.  FTL has been designed as an 
+extension to the `ICU MessageFormat`_ standard.
```
Pike: I'd drop the ICU stuff here. But add a link to the syntax doc instead?

stasm:
Thanks to you both for the comments. I addressed all feedback.  Pike already lgtm'ed this on IRC. @zbraniecki, would you like to give it another read?

zbraniecki:
r+

PR #154, opened 2016-10-12
Address feedback for the syntax.rst

zbraniecki:
@Pike , @stasm - r?

### DIFF
--- a/docs/syntax.rst
+++ b/docs/syntax.rst
@@ -2,9 +2,14 @@
 The Guide To The FTL Syntax
 ===========================
 
-FTL is an open source, localization-specific scripting language used to process 
-gender, plurals, conjugations, and most of the other quirky elements of natural 
-language.
+FTL is a localization file format used for describing translation resources.
+
+It assigns messages in a given language to unique identifiers used by
+the software to retrieve the message.
+
+FTL is designed to be simple to read, but at the same time allows to represent
+complex concepts from natural languages like gender, plurals, conjugations,
+and others.
 
 The following chapters will demonstrate how to use FTL to solve localization 
 challenges. Each chapter contains a hands-on example of simple FTL concepts.
@@ -49,7 +54,7 @@ FTL entities mostly store string values. A string is a sequence of characters
 that you can assign to an entity, store, and retrieve.
 
 By default, a string begins after a ``=`` and ends with the end of line.  You 
-can also define easy-to-read, multiline strings with a pipe operator, as can be 
+can also define easy-to-read, multiline strings with a pipe mark-up, as can be
 seen in the ``description`` entity.
 
 In almost all cases, leading and trailing spaces are not meaningful and will be 
@@ -57,6 +62,30 @@ ignored allowing you to align the string id and values in a resource file for
 better readability.  In a rare cases where leading and/or trailing spaces are 
 meaningful, FTL allows for special quote delimited strings as can be seen in 
 the ``more-info`` entity.
+
+
+Entity References
+=================
+
+::
+
+    brandName = Loki
+    installing = Installing { brandName }.
+
+    menu-save = Save
+    help-menu-save = Click "{ menu-save }" to save the file.
+
+Strings in FTL may use special syntax to incorporate small pieces of
+programmable interface. Those pieces are denoted with curly braces ``{`` and
+``}`` and are called placeables.
+
+One example of a placeable is a reference to another entity.
+
+Referencing other entities generally helps to keep certain translations
+consistent across the interface and makes maintenance easier.  It is also
+particularly handy for keeping branding separated from the rest of the
+translations, so that it can be changed easily when needed, e.g. during the
+build process of the application.
 
 
 Interpolation and External Arguments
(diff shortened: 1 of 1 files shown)


docs/syntax.rst
```diff
@@ -2,9 +2,14 @@
 The Guide To The FTL Syntax
 ===========================
 
-FTL is an open source, localization-specific scripting language used to process 
-gender, plurals, conjugations, and most of the other quirky elements of natural 
-language.
+FTL is a localization language used for describing translation resources.
```
stasm: Even simpler:

    FTL is a syntax for describing translations.

docs/syntax.rst
```diff
@@ -2,9 +2,14 @@
...
+
+It assigns messages in a given language to unique identifiers used by
+the software to retrieve the message.
+
+FTL is designed to be simple to read, but allow to represent complex concepts
```
stasm: Maybe:

    but at the same time allows to represent...

docs/syntax.rst
```diff
@@ -2,9 +2,14 @@
...
+from natural languages like gender, plurals, conjugations,
+and most of the other quirky elements.
```
stasm: Drop the last part (`and most..`) and:

    from natural languages like genders, plurals, grammatical cases, conjugations and others.

docs/syntax.rst
```diff
@@ -59,6 +64,29 @@ meaningful, FTL allows for special quote delimited strings as can be seen in
...
+
+::
+
+    brandName = Loki
+    installing = Installing { brandName }.
+
+    menu-save = Save
+    help-menu-save = Click "{ menu-save }" to save the file.
+
+In FTL strings may use special syntax to incorporate small pieces of
```
stasm: Let's use this opportunity to fix the wording here:

    Strings in FTL may use...

docs/syntax.rst
```diff
@@ -59,6 +64,29 @@ meaningful, FTL allows for special quote delimited strings as can be seen in
...
+programmable interface. Those pieces are denoted with curly braces ``{`` and
+``}`` and are called placeables.
+
+One example of a placeable is a reference to another entity.
+
+This generally helps to keep certain translations consistent across the
```
stasm: Replace `this` with `Referencing other entities...`.

docs/syntax.rst
```diff
@@ -74,10 +102,9 @@ Interpolation and External Arguments
         "emailCount": 5
     }
 
-In FTL strings may use special syntax to incorporate small pieces of 
-programmable interface. Those pieces are denoted with curly braces ``{`` and 
-``}`` and are called placeables.  The most common use case for a placeable is 
-to put an external argument, provided by the developer, into the string.
+
+The most common use case for a placeable is to put an external argument,
```
stasm: Since we've just mentioned entity references:

    `Another common use case...`

docs/syntax.rst
zbraniecki: I prefer what I wrote over this. I took it from Wikipedia articles for HTML and CSS and I believe telling people that it's a language is more appropriate than that it's a syntax.

stasm: OK, I'm not going to block on this but keep in mind that in our domain the word `language` is ambiguous and there's a good chance people will think about the natural language here.  How about `file format`?

zbraniecki: oh, good call! changing :)

Pike:
The whitespace handling still sounds undefined, which it probably isn't.

I think the paragraph about contexts can go away.

The usage of "assign, retrieve, store" feels pretty heavy. Can we get by without any of them?

zbraniecki:
>  The whitespace handling still sounds undefined, which it probably isn't.

Which paragraph it is?
The only paragraph related to white spaces I see is the `Working with text` and it describes what we do with white spaces. Not sure what's undefined about it.

> I think the paragraph about contexts can go away.

There is one in `Interpolation and External Arguments` and one in `Hello World`.
Which one?


> The usage of "assign, retrieve, store" feels pretty heavy. Can we get by without any of them?

There are:
 * 2 `assign`
 * 2 `retrieve`
 * 8 `store`

I'll fix the stores, are you ok with the volume of assign and retrieve?

Pike:
Whitespace is the last paragraph in https://github.com/zbraniecki/l20n.js/blob/256e3991fcf659a955b262ebb65c1c18707b5f3c/docs/syntax.rst#working-with-text-multiline-quote-delimited-strings.

Context paragraph is the last one in https://github.com/zbraniecki/l20n.js/blob/256e3991fcf659a955b262ebb65c1c18707b5f3c/docs/syntax.rst#hello-world.

Thanks for fixing the stores. I'd also look at the assign/retrieve usages, just because I think they're opportunities to use easier language. For example

> FTL is a localization file format used for describing translation resources.
> It assigns messages in a given language to unique identifiers used by the software to retrieve the message.

vs

> FTL is a localization file format used for describing translation resources.
> Software refers to the messages in a given language through unique identifiers.

... though, thinking about that sentence, we should add something that explains that we're doing mono-lingual files?

PR #155, opened 2016-10-13
Add a test to Karma that tests whole cycle of dom l10n

zbraniecki:
@stasm - r?

### DIFF
--- a/test/bindings/dom.js
+++ b/test/bindings/dom.js
@@ -1,32 +1,37 @@
 'use strict';
 
+function ftl(strings) {
+  const [code] = strings;
+  return code.replace(/^\s*/mg, '');
+}
+
+
 /* global window, document, assert */
 
-var dom = L20n.dom;
 describe('getResourceLinks', function() {
   it('should return an empty array if no l10n links in head', function() {
-    var head = document.createElement('head');
+    const head = document.createElement('head');
 
-    var result = L20n.getResourceLinks(head);
+    const result = L20n.getResourceLinks(head);
 
     assert.ok(result.constructor === Map);
     assert.equal(result.size, 0);
   });
 
   it('should collect all l10n links from head', function() {
-    var links = [
+    const links = [
       'locales/test.{locale}.l20n',
       '/shared/date/date.{locale}.l20n'];
-      var head = document.createElement('head');
+      const head = document.createElement('head');
 
-      for (var i in links) {
-        var link = document.createElement('link');
+      for (const i in links) {
+        const link = document.createElement('link');
         link.setAttribute('rel', 'localization');
         link.setAttribute('href', links[i]);
         head.appendChild(link);
       }
 
-      var result = L20n.getResourceLinks(head);
+      const result = L20n.getResourceLinks(head);
 
       assert.equal(result.size, 1);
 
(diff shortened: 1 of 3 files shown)


test/bindings/dom.js
```diff
@@ -80,3 +83,96 @@ describe('getAttributes', function() {
...
+key3 = Value 3
+      `
+      });
+      return Promise.resolve([
+        rb
+      ]);
+    }
+
+    function createContext(lang) {
+      return new L20n.MessageContext();
```
stasm: Pass `lang` into the constructor?

test/bindings/dom.js
```diff
@@ -80,3 +83,96 @@ describe('getAttributes', function() {
     assert.deepEqual(ret.args, l10nArgs);
   });
 }); 
+
+describe.skip('translateFragment', function() {
```
stasm: Do you plan to un-`skip` it?

test/bindings/dom.js
```diff
@@ -80,3 +83,96 @@ describe('getAttributes', function() {
...
+      ]);
+    }
+
+    function createContext(lang) {
+      return new L20n.MessageContext();
+    }
+    const loc = new L20n.HTMLLocalization(requestBundles, createContext);
+    dom.set('main', loc);
+    return dom.translateFragment(div).then(() => {
+      dom.localizations.delete('main');
```
stasm: Why delete it?

test/bindings/dom.js
```diff
@@ -80,3 +83,96 @@ describe('getAttributes', function() {
...
+      assert.equal(div.getElementsByTagName('p')[0].textContent, 'Value 1');
+      assert.equal(div.getElementsByTagName('p')[1].textContent, 'Value 2');
+      assert.equal(div.getElementsByTagName('p')[2].textContent, 'Value 3');
+    });
```
stasm: Should there be a `.then(done, done)` here?

test/bindings/dom.js
```diff
@@ -80,3 +83,96 @@ describe('getAttributes', function() {
...
+
+    function createContext(lang) {
+      return new L20n.MessageContext();
+    }
+    const loc = new L20n.HTMLLocalization(requestBundles, createContext);
+    dom.set('main', loc);
+
+    dom.observeRoot(div);
+
+    mo.observe(div, {
```
stasm: Instead of creating a new `MutationObserver` here, maybe we should consider emitting events on `document` when mutation records are translated?  I think we'll end up having to do it anyways so this might be a good opportunity to add this and make testing easier.  I can file a bug for this.  Or, if you prefer, we can land this test as-is and then change it.

test/bindings/dom.js
```diff
@@ -38,13 +37,15 @@ describe('getResourceLinks', function() {
 
 describe('setAttributes', function() {
   it('should set id on an element', function() {
+    let dom = new L20n.LocalizationObserver();
```
zbraniecki: To isolate the behavior and avoid interference between tests

test/bindings/dom.js
zbraniecki: yes :)

test/bindings/dom.js
zbraniecki: to clean up after the test.

test/bindings/dom.js
zbraniecki: So, I'm generally not in favor of emitting event after each mutation record translation. I know you are.
I believe that MO is a perfect way to observe changes to an element and that's what we're observing here :)

test/bindings/dom.js
stasm: I think it good to either clean up _or_ recreate instances for each test, but doing both is probably overkill and even a bit confusing :)

zbraniecki: fair, I prefer to make sure we use new instance each time :)

test/bindings/dom.js
```diff
@@ -80,3 +83,96 @@ describe('getAttributes', function() {
...
+      <div>
+        <p data-l10n-id="key3"></p>
+      </div>
+    `;
+
+    function requestBundles() {
+      let rb = new L20n.ResourceBundle('en-US', [
+        './locales/test.ftl'
+      ], {
+        './locales/test.ftl': `
```
zbraniecki: can't import, but can copy

PR #156, opened 2016-10-14
Bug 1308893 - Move overlay logic into bindings

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1308893


docs/bindings.md
```diff
@@ -0,0 +1,220 @@
...
+Set the `data-l10n-id` and `data-l10n-args` attributes on DOM elements.
+L20n makes use of mutation observers to detect changes to `data-l10n-*`
+attributes and translate elements asynchronously.  `setAttributes` is
+a convenience method which allows to translate DOM elements declaratively.
+
+You should always prefer to use `data-l10n-id` on elements (statically in
+HTML or dynamically via `setAttributes`) over manually retrieving
+translations with `format`.  The use of attributes ensures that the
+elements can be retranslated when the user changes their language
+preferences.
```
zbraniecki: I'd like to suggest explicitly saying why we want users to set the attribute instead of just listing one example of how we may use it:

`Setting the attribute establishes a binding between the element and the entity which L20n can use for, among others, retranslation when teh user changes their language preferences`

docs/messagecontext.md
```diff
@@ -40,6 +40,8 @@ Returns **[string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Refer
 
 # FTLList
 
+# FTLList
+
```
zbraniecki: ?

src/bindings/overlay.js
```diff
@@ -95,13 +130,80 @@ function overlay(l10n, sourceElement, translationElement) {
...
+ * @returns {boolean}
+ * @private
+ */
+function isElementAllowed(element) {
+  const allowed = ALLOWED_ELEMENTS[element.namespaceURI];
+  if (!allowed) {
+    return false;
+  }
+
+  return allowed.indexOf(element.tagName.toLowerCase()) !== -1;
```
zbraniecki: `includes`

src/bindings/overlay.js
```diff
@@ -95,13 +130,80 @@ function overlay(l10n, sourceElement, translationElement) {
...
+  const allowed = ALLOWED_ATTRIBUTES[element.namespaceURI];
+  if (!allowed) {
+    return false;
+  }
+
+  const attrName = attr.name.toLowerCase();
+  const elemName = element.tagName.toLowerCase();
+
+  // Is it a globally safe attribute?
+  if (allowed.global.indexOf(attrName) !== -1) {
```
zbraniecki: `includes`

src/bindings/overlay.js
```diff
@@ -95,13 +130,80 @@ function overlay(l10n, sourceElement, translationElement) {
...
+    return true;
+  }
+
+  // Are there no allowed attributes for this element?
+  if (!allowed[elemName]) {
+    return false;
+  }
+
+  // Is it allowed on this element?
+  if (allowed[elemName].indexOf(attrName) !== -1) {
```
zbraniecki: `includes`

docs/bindings.md
stasm: I'd be afraid of using words like `binding` here.  It's not like we're creating some mapping in memory for this.  I 'd like to think the current wording is enough.

docs/messagecontext.md
stasm: I'd know what this is, some artifact of `make docs`.  I'd like to fix `make docs` in general so let's ignore this for now.

src/bindings/overlay.js
stasm: I don't think we can yet, sadly.  It's not well supported and IIRC Babel doesn't transpile it unless you use a special plugin and then it only work for array literals anyways (`['a', 'b'].includes('a')`).

PR #157, opened 2016-10-14
Bug 1310263 - Merge ContentLocalizationObserver and ChromeLocalizationObserver into LocalizationObserver

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1310263


src/bindings/dom.js
```diff
@@ -281,6 +281,20 @@ export default class LocalizationObserver {
...
   }
 
+  translateRootContent(root) {
+    markStart();
+
+    const anonChildren = document.getAnonymousNodes ?
+      document.getAnonymousNodes(root) : null;
+    if (!anonChildren) {
+      return this.translateFragment(root).then(markEnd);
+    }
```
zbraniecki:
```javascript
if (document.getAnonymousNodes) {
  return this.translateFragment(
    document.getAnonymousNodes(root)
  ).then(markEnd);
}
```

src/bindings/dom.js
```diff
@@ -281,6 +281,20 @@ export default class LocalizationObserver {
...
+
+    return Promise.all(
+      [root, ...anonChildren].map(node => this.translateFragment(node))
+    ).then(markEnd);
+  }
```
zbraniecki: and then you can remove anonChildren from here.

src/bindings/dom.js
stasm: I think there was a reason why we wanted to translate both the anonymous content attached to the root _as well as_ the root itself.  I don't remember exactly right now btu if we want to change the logic here I'd suggest filing a separate bug.

zbraniecki: oh, sorry, I mean `if (!document.getAnonymousNodes) ...`

PR #158, opened 2016-10-14
Address feedback on gecko.rst

zbraniecki:
@Pike, @stasm: r?

### DIFF
--- a/docs/gecko.rst
+++ b/docs/gecko.rst
@@ -35,22 +35,26 @@ L20n will show the identifier of the message.
 Create An FTL Translation Resource
 ----------------------------------
 
-Translations for L20n are stored in ``.ftl`` files.  You can find more 
-information about the FTL format at the end of this document.
+Translations for L20n are stored in ``.ftl`` files.  To learn more about
+the FTL file format, see `Syntax`_.
 
-The FTL files leverage the existing packaging infrastructure for locales.  The
-translations for ``browser`` can be found in
-``browser/locale/*/browser``, as specified by ``browser/locale/jar.mn``::
+.. _Syntax: https://github.com/l20n/l20n.js/blob/master/docs/syntax.rst
 
-    browser                                        (%browser/*.ftl)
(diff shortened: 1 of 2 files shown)


docs/gecko.rst
```diff
@@ -38,19 +38,19 @@ Create An FTL Translation Resource
 Translations for L20n are stored in ``.ftl`` files.  You can find more 
 information about the FTL format at the end of this document.
```
Pike: This still refers to the inline ftl section we removed.

Move the link up to here?

docs/gecko.rst
```diff
@@ -38,19 +38,19 @@ Create An FTL Translation Resource
...
 
-The FTL files leverage the existing packaging infrastructure for locales.  The 
-*en-US* translations for ``browser`` can be found in 
-``browser/locale/en-US/browser``, as specified by ``browser/locale/jar.mn``::
+The FTL files leverage the existing packaging infrastructure for locales.  The
+translations for ``browser`` can be found in
+``browser/locale/*/browser``, as specified by ``browser/locale/jar.mn``::
```
Pike: Jein ;-)

Packaging FTL files in Firefox is done by adding their directory to jar.mn. Please only package complete directory structures, and don't try to move files around.

The top-level directory like `browser` is close to where your code is. Inside `browser/locales/en-US`, the first directory clarifies the target audience of the feature.

docs/gecko.rst
```diff
@@ -38,19 +38,19 @@ Create An FTL Translation Resource
...
 
     browser                                        (%browser/*.ftl)
```
Pike: [localization] @AB_CD@.jar:
        browser    (%browser/**/*.ftl)

is what we use now.

docs/gecko.rst
stasm: I don't understand the `don't try to move files around` part.  Can you elaborate?

stasm: Maybe rephrase like so:

    Please only package complete directory structures, and keep 
    the same directory layout as in the source.

PR #160, opened 2016-10-19
Add builtins docs

zbraniecki:
First pass, @Pike , @stasm ?

### DIFF
--- a/docs/builtins.rst
+++ b/docs/builtins.rst
@@ -31,8 +31,9 @@ Function parameters
 ===================
 
 Functions may (but don't have to) accept positional and keyword arguments.
-Some parameters are only available to developers passing the Function as
-an argument.
+Some keyword parameters are only available to developers and cannot be used
+inside of FTL translations. Instead they can be set on external arguments
+passed into translations by the developer.
 See the reference below for more information about the arguments accepted for
 each built-in function.
 
@@ -60,7 +61,7 @@ environment.
       ''minimumSignificantDigits''
       ''maximumSignificantDigits''
 
-    Devloper parameters::
+    Developer parameters::
 
       ''style''
       ''currency''
@@ -88,7 +89,7 @@ environment.
       ''second''
       ''timeZoneName''
 
-    Devloper parameters::
+    Developer parameters::
 
       ''timeZone''
 
@@ -112,7 +113,7 @@ environment.
       ''minimumSignificantDigits''
       ''maximumSignificantDigits''
 
-    Devloper parameters::
+    Developer parameters::
 
       ''type''
 
@@ -120,7 +121,6 @@ environment.
 
 ``LIST``
     Formats a list to a string in a given locale.
-    Accepts the same parameters as `Intl.ListFormat`_ API.
 
     Example::
 
@@ -132,7 +132,7 @@ environment.
 
       ''style''
 
-    Devloper parameters::
+    Developer parameters::
 
       ''type''
 
@@ -147,16 +147,24 @@ environment.
       unread-emails = Number of unread emails: { LEN($emails) }
 
 ``TAKE``
-    Returns a slice of a list.
+    Returns a slice of a list starting from the first element.
     Similar to JS `Array.prototype.slice`_ API.
+
+    Function accepts a single positional parameter which must be a number which
+    defines the number of elements that will be returned from the array.
 
     Example::
 
-      first-user = Primarely, { TAKE($users, 1) } likesd your message.
+      first-user = Primarely, { TAKE($users, 1) } likes your message.
+
+
 
 ``DROP``
     Returns a slice of the list starting from a given index.
     Similar to JS `Array.prototype.slice`_ API.
+
+    Function accepts a single positional parameter which must be a number which
+    defines the parameter from which the returned slice begins.
 
     Example::
 
--- a/docs/builtins.rst
+++ b/docs/builtins.rst
@@ -1,53 +1,101 @@
-================
-Builtins in L20n
-================
+=================
+Functions in L20n
+=================
+
+Functions provide additional functionality available to the localizers.
+They can be either used to format data according to the current language's
+rules or can provide additional data that the localizer may use (like, the
+platform, or time of the day) to fine tune the translation.
 
 L20n ships with a number of built-in functions that can be used from within
 localization messages.
 
-The list of built-ins is extensible and environments may want to add additional
-built-ins designated to aid localizers writing translations targeted
-for a given environment.
+The list of available functions is extensible and environments may want to
+add additional functions designated to aid localizers writing translations
+targeted for a given environment.
 
 
-How to use Built-ins
+How to use Functions
 ====================
 
-Localizers can reference builtin in a message within a placeable and either
-display the return value, or use it in the selector expression to select
-the right value.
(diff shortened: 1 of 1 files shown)


docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
+================
+Builtins in L20n
```
stasm: Let's use the word `Function` to describe the general class of these things, and then `Built-in Function` to describe those that are provided by `MessageContext` by default.

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+================
+
+L20n ships with a number of built-in functions that can be used from within
+localization messages.
```
stasm: I'm missing a sentence about the purpose of `Functions`.  Something about formatting data according to the current language's rules would be enough.

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+
+The list of built-ins is extensible and environments may want to add additional
```
stasm: `The list of built-in functions is…`

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+
+Default built-ins
+=================
+
+Default built-ins are very generic and should be applicable to any translation
+environment.
+
+``NUMBER``
+    Formats a number to a string in a given locale.
+    Accepts the same parameters as `Intl.NumberFormat`_ API.
```
stasm: I know it's repetition but I think we should provide a quick summary of the available parameters in form of a table for each function.

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+built-ins designated to aid localizers writing translations targeted
+for a given environment.
+
+
+How to use Built-ins
+====================
+
+Localizers can reference builtin in a message within a placeable and either
+display the return value, or use it in the selector expression to select
+the right value.
```
stasm: `FTL Functions can only be called inside of placeables. Use them to return a value to be interpolated in the message or as selectors in select expressions.  Functions may (but don't have to) accept positional and keyword arguments. See the reference below for more information about the arguments accepted for each built-in function.`

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+
+      users = { LIST($users) }
+
+      users2 = { $users }
+
+
+Partially resolved built-ins
+============================
+
+In the future we'll want to allow for partially resolved builtins to be
```
stasm: That's already possible right now! https://github.com/l20n/l20n.js/blob/3ba0b9ad8f20a9540eb7ef6c4bda4604af4677e1/src/intl/polyfill.js#L6-L7

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+constructed by the developer and passed to the localization.
+This scenario will be used when the developer wants to define the
+default parameters for the built-in, but allow the localizer
+to override them if needed.
+
+It may look like this::
+
+  main.js:
+
+  let s = document.l10n.formatValue('key1', {
```
stasm: Let's keep this doc scoped to FTL and the `MessageContext` class.  Change this exampe to use `ctx.format()`?

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+  key1 = Current date is { $date }
+
+  # If the localizer needs to change the way we display month to fit the text
+  key1 = Current date is { DATETIME($date, month=short) }
+
+
+Gecko runtime specific built-ins
+================================
+
+At the moment Gecko supports the following OS-specific built-ins:
```
stasm: Should this go into `gecko.rst`?

stasm: Also, I'm thinking we should rename this function to `PLATFORM` and return full names: `windows`, 'macos`, 'gnome', 'unix`.

docs/builtins.rst
```diff
@@ -0,0 +1,217 @@
...
+====================
+
+Localizers can reference builtin in a message within a placeable and either
+display the return value, or use it in the selector expression to select
+the right value.
+
+Examples::
+
+
+  today-is = Today is { DATETIME($date, weekday=long) }
```
stasm: Our current syntax requires `:` for the keyword arguments and quotes:

    today-is = Today is { DATETIME($date, weekday: "long") } 

https://github.com/l20n/l20n.js/blob/3ba0b9ad8f20a9540eb7ef6c4bda4604af4677e1/src/ftl/ast/parser.js#L503

We should probably change the EBNF grammar: https://github.com/l20n/spec/blob/126b422ae146cdb81eed38484eca8d732cec46a6/grammar.ebnf#L49

docs/builtins.rst
zbraniecki: Not sure. @Pike ?

stasm: I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1311666 to track the renaming.  Let's not block on it.

docs/builtins.rst
```diff
@@ -0,0 +1,296 @@
...
+
+      ''currencyDisplay''
+      ''useGrouping''
+      ''minimumIntegerDigits''
+      ''minimumFractionDigits''
+      ''maximumFractionDigits''
+      ''minimumSignificantDigits''
+      ''maximumSignificantDigits''
+
+    Devloper parameters::
```
stasm: `Developer` (typo here and elsewhere).

docs/builtins.rst
```diff
@@ -0,0 +1,296 @@
...
+
+
+  today-is = Today is { DATETIME($date) }
+
+
+Function parameters
+===================
+
+Functions may (but don't have to) accept positional and keyword arguments.
+Some parameters are only available to developers passing the Function as
```
stasm: Maybe: `Some keyword parameters are only available to developers and cannot be used inside of FTL translations.  Instead they can be set on external arguments passed into translations by the developer.`

docs/builtins.rst
```diff
@@ -0,0 +1,296 @@
...
+
+    Devloper parameters::
+
+      ''type''
+
+    See the `Intl.PluralRules`_ for the description of the parameters.
+
+``LIST``
+    Formats a list to a string in a given locale.
+    Accepts the same parameters as `Intl.ListFormat`_ API.
```
stasm: Remove as line 139 has the same info.

docs/builtins.rst
```diff
@@ -0,0 +1,296 @@
...
+``LEN``
+    Returns the number that represents the length of the list argument.
+    Similar to JS `Array.length`_ API.
+
+    Example::
+
+      unread-emails = Number of unread emails: { LEN($emails) }
+
+``TAKE``
+    Returns a slice of a list.
```
stasm: `... starting from the first element`.  Please describe the parameters that `TAKE` and `DROP` accept.

docs/builtins.rst
```diff
@@ -0,0 +1,304 @@
...
+  key1 = Today is { DATETIME($day, weekday: "short") }
+
+
+
+Gecko runtime specific functions
+================================
+
+At the moment Gecko runtime adds the following functions:
+
+``PLATFORM``
```
stasm: The bug about the renaming hasn't yet been resolved, so it's up to you if you want to use the new name here or the old one.

PR #161, opened 2016-10-20
Bug 1311662 - Remove the PLURAL builtin function

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1311662

### DIFF
--- a/docs/syntax.rst
+++ b/docs/syntax.rst
@@ -127,21 +127,20 @@ Builtins
         "unreadEmails": 5
     }
 
-In some rare cases the data provided by the developer will require some 
-additional formatting before it can be placed into the string.  FTL provides 
-a list of built-in functions that can help with common operations on the 
+In some rare cases the data provided by the developer will require some
+additional formatting before it can be placed into the string.  FTL provides
+a list of built-in functions that can help with common operations on the
 external arguments.
 
-By default, FTL can guess which formatter to run on each kind of argument: 
-``DATE``, ``NUMBER``, ``LIST`` etc., but you can also call the builtin 
+By default, FTL can guess which formatter to run on each kind of argument:
+``DATE``, ``NUMBER``, ``LIST`` etc., but you can also call the builtin
 explicitly.
 
-Explicit calls are useful because they allow you to pass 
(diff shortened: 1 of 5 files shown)


docs/syntax.rst
```diff
@@ -164,12 +164,11 @@ One of the most common cases when a localizer needs to use a placeable is when
...
-FTL provides a select expression that chooses one of the provided variants 
-based on the given selector.
+FTL provides the select expression syntax, which chooses one of the provided
+variants based on the given selector.
 
-By default, when a number is used as a selector, FTL implicitly uses ``PLURAL`` 
-formatter that selects the proper plural case for a given language. In English 
-it will be either ``one`` or ``other``.
+When the selector is a number FTL implicitly converts it into the proper plural 
+case for a given language. In English it will be either ``one`` or ``other``.
```
Pike: I'd add an explanation on when to use NUMBER($foo, {}). I.e., if you specify a particular number format in the placables, you should use the same number format in the plural rule, so that things like ordinals and trailing digits are handled correctly.

docs/syntax.rst
```diff
@@ -160,16 +159,26 @@ Selectors
...
-it will be either ``one`` or ``other``.
+One of the most common cases when a localizer needs to use a placeable is when
+there are multiple variants of the string that depend on some external
+argument.  FTL provides the select expression syntax, which chooses one of the
+provided variants based on the given selector.
+
+The selector may be a string in which case it will be compared directly to the
+keys of variants defined in the select expression.  Or it can be a number, like
+in the example above.  FTL will implicitly convert it into the proper plural
+case for a given language. In English it will be either ``one`` or ``other``.
```
Pike: I actually think we can be better on the "implicitly convert". Something like:

For numbers, the variants either match numbers exactly, or plural form names.

Not really happy with the nomenclature here, but I think we should make clear that it's both, and not just one or other.

Also, should we make it explicit that this is CLDR, and what the maximum set of options are?

stasm: Agree on both accounts, I'll update the doc.

PR #162, opened 2016-10-28
Bug 1309564 - Make bindings only set attributes with matching namespace.

zbraniecki:


src/bindings/overlay.js
```diff
@@ -130,7 +141,10 @@ function overlay(sourceElement, translationElement) {
   // cleared if a new language doesn't use them; https://bugzil.la/922577
   if (translationElement.attributes) {
     for (k = 0, attr; (attr = translationElement.attributes[k]); k++) {
-      if (isAttrAllowed(attr, sourceElement)) {
+      if (isAttrAllowed({
```
stasm: `isAttrAllowed` is meant to accept a DOM Attribute-like object, similar to `isElementAllowed` accepting a DOM Element.  I'd like to keep it that way, so I'd suggest to just pass `attr` here, which already has the `name` and the `namespaceURI` properties.

Also, this code shouldn't just take `translationElement`'s namespace as the attribute's namespace. Given an XHTML document and a translation value of `foo <span xml:lang="en">bar</span>!`, the `span` element's namespace is `http://www.w3.org/1999/xhtml` but the `lang`'s attribute namespace is `http://www.w3.org/XML/1998/namespace`.

src/bindings/overlay.js
```diff
@@ -130,7 +141,10 @@ function overlay(sourceElement, translationElement) {
...
+          ns: DOM_NAMESPACES[translationElement.namespaceURI],
+          name: attr.name
+        }, sourceElement)) {
         sourceElement.setAttribute(attr.name, attr.value);
```
stasm: We might want to start using `setAttributeNS` as well as `attr.localName` throughout this file.  We can do it in a follow up.

src/bindings/overlay.js
```diff
@@ -169,6 +183,10 @@ function isElementAllowed(element) {
  * @private
  */
 function isAttrAllowed(attr, element) {
+  // Does it have a namespace that matches the element's?
+  if (attr.ns === null || DOM_NAMESPACES[attr.ns] !== element.namespaceURI) {
```
stasm: If you pass the DOM Attribute object as `attr` here, you can simply do:

    if (attr.namespaceURI !== element.namespaceURI)

src/bindings/overlay.js
```diff
@@ -37,6 +37,15 @@ const ALLOWED_ATTRIBUTES = {
   }
 };
 
+const DOM_NAMESPACES = {
+  'html': 'http://www.w3.org/1999/xhtml',
+  'xul': 'http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul',
+
+  // Reverse map for overlays.
+  'http://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul': 'xul',
+  'http://www.w3.org/1999/xhtml': 'html',
```
stasm: The reverse map shouldn't be required if you pass actual DOM Attribute objects or `{ name, namespaceURI }` objects to `toAttrAllowed`.

src/lib/format.js
```diff
@@ -138,7 +138,9 @@ export function entityFromContext(ctx, errors, id, args) {
     for (let i = 0, trait; (trait = entity.traits[i]); i++) {
       const attr = ctx.format(trait, args, errors);
       if (attr !== null) {
-        formatted.attrs[trait.key.name] = attr;
+        let key =
+          trait.key.ns ? `${trait.key.ns}/${trait.key.name}` : trait.key.name;
+        formatted.attrs[key] = attr;
```
stasm: This looks like a step backwards.  First the parser parses the namespaces and then this code serializes them back, just to `split('/')` again in `overlayElement`.

I'd prefer something more semantic:

 - either store namespaced attributes in `formatted[Symbol.for(trait.key.ns)][trait.key.name]`, or
 - make `formatted.attrs ` an array and push `[namespace, key, value]` tuples to it.

Also, we should ignore all traits which don't have any namespaces here as they are local to the language and shouldn't be exposed to the app.  You could check for `trait.key.ns` even before calling `ctx.format`.

src/bindings/overlay.js
zbraniecki: II couldn't go for it because HTML doesn't have attribute namespaces.

On IRC we agreed to use target element's namespace

src/lib/format.js
zbraniecki: went for a list to minimize the structure complexity

src/lib/format.js
```diff
@@ -134,11 +134,15 @@ export function entityFromContext(ctx, errors, id, args) {
   };
 
   if (entity.traits) {
-    formatted.attrs = Object.create(null);
+    formatted.attrs = [];
     for (let i = 0, trait; (trait = entity.traits[i]); i++) {
       const attr = ctx.format(trait, args, errors);
       if (attr !== null) {
-        formatted.attrs[trait.key.name] = attr;
+        formatted.attrs.push([
```
stasm: Would it make sense to only format traits where `trait.key.ns` is truthy?

src/bindings/overlay.js
```diff
@@ -168,7 +178,11 @@ function isElementAllowed(element) {
  * @returns {boolean}
  * @private
  */
-function isAttrAllowed(attr, element) {
+function isAttrAllowed(ns, attr, element) {
+  // Does it have a namespace that matches the element's?
+  if (ns !== element.namespaceURI) {
+    return false;
```
stasm: This looks like an unnecessary factorization to make the code in line 77 _and_ line 143 both work.  In case of line 143 this check will always pass.  In line 77 you could simply filter non-matching attributes earlier.

src/bindings/overlay.js
```diff
@@ -63,9 +68,14 @@ export default function overlayElement(element, translation) {
...
-  for (const key in translation.attrs) {
-    if (isAttrAllowed({ name: key }, element)) {
-      element.setAttribute(key, translation.attrs[key]);
+  if (translation.attrs === null) {
+    return;
+  }
+
+  for (const [ns, name, value] of translation.attrs) {
+    if (ns !== undefined &&
+        isAttrAllowed(DOM_NAMESPACES[ns], { name }, element)) {
```
stasm: How about:

    if (DOM_NAMESPACES[ns] === element.namespaceURI &&
        isAttrAllowed({ name }, element) {

src/bindings/overlay.js
```diff
@@ -63,9 +68,14 @@ export default function overlayElement(element, translation) {
...
   }
 
-  for (const key in translation.attrs) {
-    if (isAttrAllowed({ name: key }, element)) {
-      element.setAttribute(key, translation.attrs[key]);
+  if (translation.attrs === null) {
+    return;
+  }
+
+  for (const [ns, name, value] of translation.attrs) {
```
stasm: I've had bad experiences using destructing assignment in hot code before.  You probably tested this and still went for this syntax here -- does this mean the performance is better nowadays?

src/lib/format.js
zbraniecki: Do we want to consider all non-ns traits as private? If so, then yes.

stasm: Yeah, I think that's what we wanted to do.

PR #168, opened 2016-11-14
Bug 1316083 - Refactor Localization into DOMLocalization

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1316083


src/bindings/document_localization.js
```diff
@@ -0,0 +1,249 @@
...
+    this.mutationObserver = new MutationObserver(
+      mutations => this.translateMutations(mutations)
+    );
+
+    this.observerConfig = {
+      attributes: true,
+      characterData: false,
+      childList: true,
+      subtree: true,
+      attributeFilter: ['data-l10n-id', 'data-l10n-args', 'data-l10n-bundle']
```
Pike: s/data-l10n-bundle/data-l10n-with/

stasm: Great catch, I fixed it here and in two other places (docs and tests).

PR #169, opened 2016-11-16
Bug 1317989 - Move bidi wrapping back into MessageContext

stasm:
https://bugzilla.mozilla.org/show_bug.cgi?id=1317989

### DIFF
--- a/src/intl/resolver.js
+++ b/src/intl/resolver.js
@@ -38,6 +38,11 @@ import builtins from './builtins';
 // Prevent expansion of too long placeables.
 const MAX_PLACEABLE_LENGTH = 2500;
 
+// Unicode bidi isolation characters.
+const FSI = '\u2068';
+const PDI = '\u2069';
+
+
 /**
  * Map an array of JavaScript values into FTL Values.
  *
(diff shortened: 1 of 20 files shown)


src/intl/resolver.js
```diff
@@ -329,15 +334,20 @@ function Pattern(env, ptn) {
...
             'Too many characters in placeable ' +
             `(${str.length}, max allowed is ${MAX_PLACEABLE_LENGTH})`
           )
         );
-        result += str.substr(0, MAX_PLACEABLE_LENGTH);
+        str = str.substr(0, MAX_PLACEABLE_LENGTH);
+      }
+
+      if (ctx.useIsolating) {
+        result += FSI + str + PDI;
```
zbraniecki: template string here to avoid two concats?

stasm: ok!

stasm: @zbraniecki Just to clarify: do you mean:

    result += `${FSI}${str}${PDI}`;

or:

    result = `${result}${FSI}${str}${PDI}`;

?

zbraniecki: I don't think latter has any impact, so former :)

PR #173, opened 2017-01-31
Bug 1331721 - Use ECMA 402's language negotiation

stasm:
Use the language negotiation defined in ECMA 402 9.2.7:

  https://tc39.github.io/ecma402/#sec-lookupsupportedlocales

With a small change in step 4d where instead of adding requested locale, we're
adding availableLocale to the subset. This allows us to directly use returned
subset to pool resources.

There was also some inconsistency in using Arrays vs. Sets for handling lists
of languages.  I unified the code to always use Arrays, because the ECMA
algorithm already uses Array everywhere.

### DIFF
--- a/src/intl/locale.js
+++ b/src/intl/locale.js
@@ -1,14 +1,445 @@
-export function prioritizeLocales(def, availableLangs, requested) {
-  const supportedLocales = new Set();
-  for (const lang of requested) {
-    if (availableLangs.has(lang)) {
-      supportedLocales.add(lang);
+/* eslint-disable */
+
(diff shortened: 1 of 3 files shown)


src/intl/locale.js
```diff
@@ -1,15 +1,446 @@
-export function prioritizeLocales(def, availableLangs, requested) {
-  const supportedLocales = new Set();
-  for (const lang of requested) {
-    if (availableLangs.has(lang)) {
-      supportedLocales.add(lang);
+/* eslint-disable */
```
stasm: There were over 200 linting errors related to the use of `var`, double quotes and others.  I just disabled linting this file to save some work.

PR #174, opened 2017-03-25
Use Fluent under the hood

stasm:
Let's start using `fluent.js` in l20n.js. More information about Fluent can be found at https://groups.google.com/forum/#!topic/mozilla.tools.l10n/NPmsJD4IGjQ and http://projectfluent.io/.


zbraniecki:
I'd like to question if this is reviewable... :(

Maybe we can just assume that the sum of all reviews I gave you is enough and I sr+ the architectural decision here?

stasm:
Yes, I think that's the level of review I'd like to get from you. To make sure you're okay with the architectural changes this PR introduces.

stasm:
BTW the reason why Travis fails is that I still haven't released the updated version of fluent and fluent-langneg which includes a fix to package.json which in turn make it possible to bundle them with l20n.js using rollup.

PR #178, opened 2017-05-02
Disable FSI/PDI isolating on IE11

stasm:
IE11 on Windows renders FSI and PDI as mojibakes. Let's turn isolating off when this browser is detected.

### DIFF
--- a/src/web/index.js
+++ b/src/web/index.js
@@ -14,7 +14,11 @@ import { documentReady, getResourceLinks, getMeta } from './util';
 // used to create new `MessageContext` objects for a given `lang` with selected
 // builtin functions.
 function createContext(lang) {
-  return new MessageContext(lang);
+  // IE11 renders FSI and PDI as mojibakes so we turn them off.
+  const isIE11 = navigator.userAgent.indexOf('Trident') > -1;
+  return new MessageContext(lang, {
+    useIsolating: !isIE11
+  });
 }
 
 // Called for every named Localization declared via <link name=…> elements.


src/web/index.js
```diff
@@ -14,7 +14,11 @@ import { documentReady, getResourceLinks, getMeta } from './util';
 // used to create new `MessageContext` objects for a given `lang` with selected
 // builtin functions.
 function createContext(lang) {
-  return new MessageContext(lang);
+  // IE11 renders FSI and PDI as mojibakes so we turn them off.
+  const isIE11 = navigator.userAgent.indexOf('Trident') > -1;
```
mgol: I'd change that to:
```
const isIE = !!document.documentMode;
```

The less UA-sniffing the better; UA can always change while no browser should ever implement the non-standard `document.documentMode`.

tsjensen:
There is currently a [bug](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/6831905/) in Edge as well which seems to be the same problem:
capture_001_26032018_214130

It would be really nice if we could extend this solution to cover this Edge bug, too.
Or maybe be just put an option to disable embedding marks altogether, which makes sense for applications which do not support any RTL languages. Thanks!

zbraniecki:
> Or maybe be just put an option to disable embedding marks altogether, which makes sense for applications which do not support any RTL languages. 

There's an option, set `useIsolation: false`.

Also, I'm in touch with the Edge team and apparently the bug is fixed in the upcoming Edge (part of the rs5 release). Not sure when it'll go public, but at least we know it'll be fixed now :)

tsjensen:
@zbraniecki Thanks for your answer!
Can you tell me how to set this option?

(Sorry to ask this here, if needed please point me to a better place to ask this, too. I could find no answer in the docs or Google. Thanks!)

zbraniecki:
> Can you tell me how to set this option?

It's one of the options available on the constructor of `MessageContext`.

tsjensen:
Thank you for the hint @zbraniecki!
However, construction of the `MessageContext` is [hardcoded](https://github.com/l20n/l20n.js/blob/master/src/web/index.js#L17). Now:
- Do I have to ship a patched *l20n.js*?
- Or do I have to repeat the [logic from l20n](https://github.com/l20n/l20n.js/blob/master/src/web/index.js#L37-L43) in my application, replacing the `document.l10n` reference?
- Or (last option), can I somehow use the information in [hacking.rst](https://github.com/l20n/l20n.js/blob/master/docs/hacking.rst#the-localization-observer-api) to register an "extra" to accomplish this? (Unsure if and how this could be.)

Thanks for your guidance!

zbraniecki:
@stasm - do you have any thoughts on how to help @tsjensen here? Should they be using l20n?

tsjensen:
Microsoft seems to have abandoned the planned fix ([6831905](https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/6831905/)).
How do you intend to react? Should we now get this PR merged? Or will you offer a way of passing the `useIsolation` flag to fluent.js (then each application could decide if it wants to regress bidi).
Thanks!

